PR #1, opened 2018-12-18
Use fixed dependencies

brunobesson:
@cbeauchesne how do you handle `package-lock.json`?
It is not in the `.gitignore` file, but you do not seem to keep it uptodate.


cbeauchesne:
Abslutly no idea, I'll have a look on it tomorrow

PR #19, opened 2018-12-27
Parse accents, get months, fix title

cbeauchesne:


src/components/map/BiodivInformation.vue
```diff
@@ -31,6 +31,29 @@
             }
         },
 
+        computed: {
+            description() {
+                let result = this.data.description || ''
+
+                // very simple html > text interpreter : only html entities
```
brunobesson: ?

cbeauchesne: I didn't go deep into biodiv data, just anticipate the case where description is null or undefined.

cbeauchesne: Or, if your question is about the comment : biodiv description contains html. I didn't want to use v-html + eslint exception (I didn't want to trust by default data from biodiv).

I also wanted to avoid an html sanitizer in the build, just for this point. So I use the simpliest way : consider that biodiv data contains only html entities, and use a fixed list of them.

brunobesson: Still, I think a sanitizer would be best.
Anyway, if you want to build a fixed list, I am wondering why yu have put only 2 cases, on `e` letters? There are more common cases than this, even if you consider only french.

cbeauchesne: on the short list, it's because I did not yet have a look on biodiv data to sse what is the list of character I need to include.

For the sanitizer, if biodiv data contains html tag, it will be mandatory. In this case, I will use webpack feature to include it in another bundle : sanitizers are quite big modules, for a popup that is displayed very rarely.

But if it does not contains html tags, only html entities, my opinion is that this fixed list is a good compromise, as we only have latin alphabet languages.

brunobesson: Still not convinced, but whatever :)
At least add all accentuated letters from the languages we support: à - â - ä - é - è - ê - ë - ï - î - ô - ö - ù - û - ü - ÿ - ç - æ - œ and the accentued versions for french.
 You may refer e.g. to https://fr.wikipedia.org/wiki/Diacritiques_utilis%C3%A9s_en_fran%C3%A7ais#Combinaisons (not sure it is complete though)

cbeauchesne: Obviously, it's not the good long term solution. But for the moment, it provides a quite good ratio results/efforts : searching for a good sanitizer + bundle configuration take some times, and I'm in late :) 

I'll have a look on your list.

PR #38, opened 2019-01-03
Allow devlopper to specify context messages for translations

cbeauchesne:
Allow dev to specify context for translations. For instance, "quality" means document quality in C2C, translated into "complétude" in french.

But int's also used as snow quality in one single place, which is different in french, and must be translated in "qualité".

context can be specified with two way : 

* a second argument in `$gettext()` function
* a `translate-context` attribute on node containing `v-translate` directive

### DIFF
--- a/src/js/vue-plugins/gettext-plugin.js
+++ b/src/js/vue-plugins/gettext-plugin.js
@@ -19,19 +19,33 @@ function cleanMessageId(msgid) {
     return msgid
 }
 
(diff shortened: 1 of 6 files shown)


src/js/vue-plugins/gettext-plugin.js
```diff
@@ -141,9 +155,14 @@ export default function install(Vue) {
                     }
 
                     element.dataset.msgid = cleanMessageId(element.innerText)
+
+                    let context = element.attributes.getNamedItem('translate-context')
```
brunobesson: could be a `const`

PR #39, opened 2019-01-04
Change common format from javascript to json, as it's pure data

cbeauchesne:
common.js is a pure data object.

It's better to store it as json : 

* better reuse in other langs (SS node, python)
* not in javascript stats

@brunobesson, I've also added a small fix in const/let def, eg your remark on last PR.


brunobesson:
> I've also added a small fix in const/let def

Soon, this should be handled by linting. But for this, we need to stick to either elsin or typescript

PR #41, opened 2019-01-04
migrate config module from js to typescript

cbeauchesne:
It's a looong road. First step !


brunobesson:
I see you requested my review, but I'm not sure what you expect from me? I have no specific remark concerning this.

PR #66, opened 2019-01-10
Display image size in ko

cbeauchesne:


brunobesson:
`ko` should be i18n'ed, don't you think? (the same for the label).
Also, having a more subtle file size could be nice, i.e. using mb when file is over 1 mb, with one number after comma.

cbeauchesne:
The label is done in transifex, just waiting a pull. 

ko, you're right! 

For the smarter rules, obviously. I add an issue (as it's not in the v6 => postponed)

PR #67, opened 2019-01-10
uppercaseFirstLetter filter, as CSS trick does not works on all browsers

cbeauchesne:


brunobesson:
You did not remove the corresponding CSS rule?

PR #78, opened 2019-01-10
add downlad outings button

cbeauchesne:


brunobesson:
This pull request **introduces 1 alert** when merging 115d95714d4d52bb0a925309e6175e342f38202b into 3a2d0373a28dfea5eb0f62d105cf02166df4303e - [view on LGTM.com](https://lgtm.com/projects/g/c2corg/c2c_ui/rev/pr-2033e464bd7d28ce5bde5b25ee41e2a77b610152)

**new alerts:**

* 1 for Incomplete string escaping or encoding

---

*Comment posted by [LGTM.com](https://lgtm.com)*

cbeauchesne:
As far as I understand the point, LGTM seems not to like when the second part of a replace contains a double anti-slash.

@brunobesson do you see something I've missed ?

brunobesson:
No, this is sort of a false positive. The rule here tries to prevent SQL injection when one uses simple replace methods. This is not pertinent here, we do no need to escape backslashes.

PR #96, opened 2019-01-13
if delete button is invisible : can't click

cbeauchesne:


src/components/generics/buttons/DeleteButton.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 
-    <fa-layers class="fa-lg" @click="$emit('click')" :class="{'hidden-button':!visible}">
+    <fa-layers class="fa-lg" @click="visible ? $emit('click') : undefined" :class="{'hidden-button':!visible}">
```
brunobesson: not fluent with the vue ternary operator, but can't the last part be omitted?

PR #109, opened 2019-01-15
Social networks sharing

cbeauchesne:


src/views/document/utils/SocialNetworkSharing.vue
```diff
@@ -0,0 +1,46 @@
...
+        class="social-network-sharing"
+        addthis:ui_click="true"
+        v-bind="$attrs"
+        :title="$gettext('Share')">
+        <fa-icon icon="share-square" />
+    </span>
+</template>
+
+<script>
+    const publicId = 'ra-58abf6b4f3a680cb'
```
brunobesson: shouldn't this be stored in a config file instead?

src/views/SideMenu.vue
```diff
@@ -28,13 +28,19 @@
 
             <div class="columns is-gapless has-text-centered is-mobile menu-socials">
                 <div class="column">
-                    <img src="@/assets/img/social/twitter.png" alt="Twitter">
+                    <a href="https://twitter.com/camptocamporg">
+                        <img src="@/assets/img/social/twitter.png" alt="Twitter">
```
brunobesson: can't we use SVG fonts instead?

cbeauchesne: We do not have a social for google+ 

That said google+ is dead. Could we see with c2c assoce to remove it ?

cbeauchesne: correction :  a svg font

brunobesson: I agree we can remove google+. But my comment was more general: you are also using images for twitter and facebook.
There are definitely fonts for brands. See e.g. https://fontawesome.com/icons?d=gallery&s=brands

cbeauchesne: yes, we still have it for twitter and fb, I will use them.

cbeauchesne: FYI : https://forum.camptocamp.org/t/projet-ui-google/235265

PR #112, opened 2019-01-15
use a datepicker, io built-in one : safari does not have one...

cbeauchesne:


src/views/wiki/edition/utils/FormInput.vue
```diff
@@ -89,11 +101,25 @@
...
 
 <script>
+
+    import Datepicker from 'vuejs-datepicker'
+    import { fr, en, es, ca, de, it, eu } from 'vuejs-datepicker/dist/locale'
+
     import { requireDocumentProperty, requireFieldProperty } from '@/js/properties-mixins'
     import InputConditionsLevels from './InputConditionsLevels'
 
+    // note that eu is missing. Sorry euskara...
```
brunobesson: have you studied how locales are built-in? it might be quite easy to add euskara and propose a patch?

cbeauchesne: No I didn't. It may be quit easy.

But considering that datepicker's translation is an evolution comparing to V6, and that euskara is a low c2c audience, it's not a big prio :)

brunobesson: Could you open a issue on c2cui anyway, to keep track of it?

cbeauchesne: #128 :)

PR #137, opened 2019-01-21
fix scroll issue

cbeauchesne:


src/views/document/utils/DocumentViewMixin.js
```diff
@@ -144,11 +144,34 @@ export default {
...
+                    .then(this.scrollToHash)
+                    .then(this.updateUrl)
+            }
+        },
+
+        scrollToHash() {
+            if (this.$route.hash) {
+                // we'll have to wait for DOM update
+                this.$nextTick(() => {
+                    let el = document.querySelector(this.$route.hash)
```
brunobesson: `el` could be a const

src/views/document/utils/DocumentViewMixin.js
```diff
@@ -144,11 +144,34 @@ export default {
...
+
+                    if (el) {
+                        let docEl = document.documentElement
```
brunobesson: const, const and const :)

src/views/document/utils/DocumentViewMixin.js
```diff
@@ -144,11 +144,34 @@ export default {
...
+                        let docRect = docEl.getBoundingClientRect()
+                        let elRect = el.getBoundingClientRect()
+                        let x = elRect.left - docRect.left
+                        let y = elRect.top - docRect.top
+                        window.scrollTo(x, y - 50) // navbar height ...
```
brunobesson: I don't understand why you are computing an `x` position: you're only trying to sroll vertically to the right anchor position?

cbeauchesne: sure :)

PR #138, opened 2019-01-21
add 300ms delay to avoid unecessary API calls

cbeauchesne:


cbeauchesne:
You're right. I've just learned what "debounce" mean :) 

I'll postpone this for the moment (except if this current code is bugged), I'd like to empty ergonomy task list, which is pretty big and scary me a little.

PR #151, opened 2019-01-23
waypoints wasn't considered on first extent

cbeauchesne:


src/components/map/OlMap.vue
```diff
@@ -600,7 +600,11 @@
                     return
                 }
 
-                var extent = this.documentsLayer.getSource().getExtent()
+                var extent = ol.extent.createEmpty()
```
brunobesson: no var :)

cbeauchesne: Thank you :) resolved in #153

PR #152, opened 2019-01-23
field unit, and field hide logic

cbeauchesne:


src/views/document/RouteView.vue
```diff
@@ -104,7 +107,7 @@
                     </markdown-section>
 
                     <markdown-section :document="document" :field="fields.gear">
-                        <div class="content" slot="after" v-if="gear_articles">
+                        <div class="content" slot="after" v-if="Object.keys(gear_articles).length !== 0">
```
brunobesson: IMHO this logic should be in the controller, not in the view

cbeauchesne:
Basiclly, I should have returned null in gear_article computation, instead of empty object. But as `v-if` is in both situation required, not such a big deal. 

For a better clarty, I should have done a isEmpty() function...

PR #157, opened 2019-01-24
good name for field description of WP

cbeauchesne:


cbeauchesne:
Codacy fails again...

PR #176, opened 2019-02-05
Respecter cest proteger

brunobesson:
Integrates respecter c'est proteger fauna protection layers into c2c.

This is not a ready PR as it, some things still need to be done:
* i18n (I have updated the `.pot` file, is there something else to do?)
* L&F needs some love (but this applies to the whole project: does it need to be done now or later?)

Still, I am willing to get some feedback now. I'll handle it when I am back from vacation 🙂


cbeauchesne:
Ok, I'm also pretty unavailable for few days, I will do asap.

cbeauchesne:
First point : there is a label visible on hovering an area : it has been removed

src/components/map/OlMap.vue
```diff
@@ -27,6 +27,14 @@
                                 {{ layer.get('title') }}
                             </div>
                         </td>
+                        <td v-if="showProtectionAreas && !editable">
```
cbeauchesne: Ergo : this check box and its title should be probably be placed above slopes part : try to the map on a route on a medium size screen : it is not visible as now.

capture d ecran 2019-02-10 a 10 35 42

src/components/map/OlMap.vue
```diff
@@ -299,11 +317,14 @@
                     new ol.control.Attribution()
                 ],
 
-                layers: this.mapLayers.concat(this.dataLayers).concat([
-                    this.biodivLayer,
+                layers: [
```
cbeauchesne: Good syntax, thank you. I'm definitlty a rookie :)

src/components/map/OlMap.vue
```diff
@@ -666,6 +689,33 @@
...
+                        )
+                        respecterCestProtegerService.identify(
+                            position,
+                            rcpExtent,
+                            this.map.getSize()[0],
+                            this.map.getSize()[1],
+                            this.$language.current
+                        ).then(({ data }) => {
+                            if (data.results && data.results.length > 0) {
+                                this.swissProtectionAreaData = data.results[0]
```
cbeauchesne: We only display item #0, not  other ones?

src/components/map/mapUtils.js
```diff
@@ -31,18 +31,15 @@ const buildTextStyle = function(title, highlight) {
     return text
 }
 
-export const buildPolygonStyle = function(title, highlight) {
-    const opacityFactor = highlight ? 1.5 : 1
```
cbeauchesne: This was used for hovering effect, why removing it?

src/js/apis/RespecterCestProtegerService.js
```diff
@@ -0,0 +1,26 @@
...
+function RespecterCestProtegerService() {
+    this.axios = axios.create()
+}
+
+RespecterCestProtegerService.prototype.identify = function(position, extent, mapWidth, mapHeight, language) {
+    if (language !== 'fr' && language !== 'de' && language !== 'it') {
+        language = 'en'
+    }
+
+    const baseUrl = 'https://api3.geo.admin.ch/rest/services/api/MapServer/identify';
```
cbeauchesne: semi ;) 

By the way, it's probably the good moment to switch to good semi config + 2 spaces indents. Let me know your opinion.

brunobesson:
> What is L&F ?

Look & feel 🙂 
So yes, your remarks concerning the rendering are true, but I haven't done anything for this now

src/components/map/OlMap.vue
brunobesson: Yes, one of the L&F problems, not dealt yet

src/components/map/OlMap.vue
brunobesson: Nah, it's only some syntaxic sugar allowed with es6. But I like it because it is more consise.

src/components/map/mapUtils.js
brunobesson: In fact, I have changed the rendering of biodivsport data to be consistent with RcP:

* no hovering effect (and no text)
* yellow area instead of a blue one to match recommended protection zones.

Maybe I should post a screenshot of the rendering?

src/js/apis/RespecterCestProtegerService.js
brunobesson: Right 👍 
As for switching to mandatory semi colon and 2 spaces indent, do it whenever you are ready for it. I can quite easily update my PR to match this (there are already conflicts anyway)

src/components/map/mapUtils.js
cbeauchesne: I've seen it, don't worry.

If you change the ergo of these areas, I recommend a small message on forum

src/js/apis/RespecterCestProtegerService.js
cbeauchesne: ok

src/components/map/mapUtils.js
brunobesson: I did the ergo the first time, nobody commented on it. I assume this change won't bother anyone 🙂 
For consistency, I cannot imagin another possibility right now.

cbeauchesne: ok, let's go!

brunobesson:
Your review still requires changes.
Also, what do we do with the map choice layout? Do we leave it as is for now?

brunobesson:
OK, will resolve conflicts and merge after your syntax PR

src/components/map/OlMap.vue
brunobesson: We assume that there are no overlapping areas. There may be several results because the API has a tolerance, but in this case, it seems legit to consider the first result returned.

brunobesson:
Replaced by new branch (due to rebasing issues)

PR #189, opened 2019-02-10
link to google captcha service

cbeauchesne:


src/views/user/LoginView.vue
```diff
@@ -216,6 +256,13 @@
             })
         },
 
+        mounted() {
+            let recaptchaScript = document.createElement('script')
```
brunobesson: could be a `const` (but don't correct this now, sometime soon we'll have auto linting for this)

vue.config.js
```diff
@@ -52,15 +52,17 @@ const config = {
...
         },
         prod: {
             name: 'prod',
             api: 'https://api.camptocamp.org',
             media: 'https://media.camptocamp.org/c2corg_active',
             imageBackend: 'https://images.camptocamp.org',
             forum: 'https://forum.camptocamp.org',
-            readWrite: false // conservative approach : you have to explicity set it to true
+            readWrite: false, // conservative approach : you have to explicity set it to true
+            recaptchaKey: '6Lc9Cw4UAAAAAIKnlar0AOsGX_P5S-bk9u8viuo2'
```
brunobesson: is it OK to store this key on the repo? it doesn't need to be hidden?

cbeauchesne: yes, it's the public key send from UI to google server (that the way I've getted them :) )

src/views/user/LoginView.vue
cbeauchesne: 👍

PR #192, opened 2019-02-11
upload image sequentially

cbeauchesne:


src/js/Worker.js
```diff
@@ -0,0 +1,30 @@
+// This class is a simple worker.
```
brunobesson: Why are you using a class now?
Up to now you always used prototype declarations.
Because it's a copy & paste?

cbeauchesne: No, it's because I now use VS code, and it proposes me this transformation out of the box. I was amazed by the feature, and I can't rollback it 😄

PR #202, opened 2019-02-13
Eslint conf

cbeauchesne:
here we go!

* HTML and JS -> two spaces indent
* JS -> semi !


.editorconfig
```diff
@@ -8,8 +8,8 @@ indent_style = spaces
 trim_trailing_whitespace = true
 
 [*.html]
-indent_size = 4
+indent_size = 2
 
 [*.js,*.ts,*.json,*.xml]
 # to be modified soon :)
```
brunobesson: remove the comment, then 🙂

.eslintrc.json
```diff
@@ -36,7 +36,10 @@
     "eqeqeq": "warn",
     "camelcase": "off",
     "no-trailing-spaces": "warn",
-    "semi": "warn",
+    "semi": [
+      "warn",
```
brunobesson: for me, it should be an error

.editorconfig
cbeauchesne: Done

.eslintrc.json
cbeauchesne: done

PR #208, opened 2019-02-15
plenty of very small fixes

cbeauchesne:


src/views/Navigation.vue
```diff
@@ -92,37 +92,45 @@
 
             <router-link :to="{ name: 'profile', params:{id:$user.id} }" class="dropdown-item is-size-5">
               <fa-icon icon="user"/>
+              <span>&nbsp;</span>
```
brunobesson: do you reaaly need to include the non breakable space in a `<span>`?

cbeauchesne: Yes, thank' to HTML interpretation of spaces : 

    <fa-icon />
    <span> </span>
    <span v-translate>...</span> 

> no space rendered

    <fa-icon />
    &nbsp;
    <span v-translate>...</span> 

> way too many spaces as both nbsp and surounding spaces are rendered

PR #254, opened 2019-02-27
Configure deployment of app in a docker container (using nginx)

brunobesson:


vue.config.js
```diff
@@ -119,6 +119,8 @@ if (process.env.BUILD_ENV === 'local' || process.env.BUILD_ENV === undefined) {
 } else if (process.env.BUILD_ENV === 'camptocamp') {
   config.urls = config.urlsConfigurations.prod;
   config.isProduction = true; // explicit prod flag
+  // set a warning if bundle size is too big
+  result.configureWebpack.performance.hints = 'warning';
```
cbeauchesne: @brunobesson is it yet necessary ? I've increased the limits, it should pass like a charm.

brunobesson: Nope. Only forgot you had already made the changes. Feel free to restore the correct behaviour!

cbeauchesne: Ok, will do it on next PR

PR #265, opened 2019-03-01
small fixes on design

flob38:
Add class has-rounded-corner : to put border-radius to 5px on title on the topoguide view.
Modify line-height on title on the topoguide view.
Modify the variable body-background-color to the forum's background color.


cbeauchesne:
Champagne :) 👍

PR #286, opened 2019-03-07
fixe flex problem when printing from firefox

flob38:
Firefox is printing only the first page of a document if the document is displayed "flex".
This bug is reported several times : 
https://bugzilla.mozilla.org/show_bug.cgi?id=1241323
https://bugzilla.mozilla.org/show_bug.cgi?id=1421637 and more.

Also modify some view to display the column on 100% width.

### DIFF
--- a/src/views/document/OutingView.vue
+++ b/src/views/document/OutingView.vue
@@ -11,7 +11,7 @@
         <tool-box :document="document"/>
       </div>
 
-      <div class="column is-9 is-12-print">
+      <div class="column is-9 is-12-print is-block-print">
 
         <div class="box">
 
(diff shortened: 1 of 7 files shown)


src/views/document/OutingView.vue
```diff
@@ -11,7 +11,7 @@
         <tool-box :document="document"/>
       </div>
 
-      <div class="column is-9 is-12-print">
+      <div class="column is-9 is-12-print is-block-print">
```
cbeauchesne: Do you really need `is-block-print`on this one? `column` is a flex child, but does not have `display:flex`

Same question for next changes on other `div.column`

flob38: Exact, it's 2 errors, the modification is on an other line.

cbeauchesne: Also, if you could has a small comment in helper.scss above ìs-block-print` declaration, with a short explanation, and a link to bugzilla page, it would be perfect.

PR #301, opened 2019-03-11
Handle forum login/logout use case

cbeauchesne:


cbeauchesne:
@brunobesson do you have an idea about the reason of travis failure? Can I merge anyway?

cbeauchesne:
Strange bug, a dummy commit fixed it...

PR #338, opened 2019-03-20
Hide svg and print text for the orientations

flob38:
To save place on printed topo, we remove the svg representing the wind rose and write the different orientations in plain text.
Don't know if it's the best way to do it...

### DIFF
--- a/src/components/generics/inputs/InputOrientation.vue
+++ b/src/components/generics/inputs/InputOrientation.vue
@@ -1,5 +1,14 @@
 <template>
-  <div class="control">
+  <div
+    class="control"
+    :title="((value_.indexOf('E')>-1)?'E ':'')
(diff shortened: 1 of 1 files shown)


src/components/generics/inputs/InputOrientation.vue
```diff
@@ -1,5 +1,14 @@
 <template>
-  <div class="control">
+  <div
+    class="control"
+    :title="((value_.indexOf('E')>-1)?'E ':'')
```
cbeauchesne: `value_` is an array, and method join will work like a charm : 

`:title="value_.join(', ')"`

And joining with comma+space will display orientation like this : 

`N, NE, E`

flob38: Yes, it's a prettiest method...

PR #339, opened 2019-03-21
Mobile cards optimisation

brunobesson:
As there is only one card per line on mobile, there is no need to limit content as much as in other situations.

* No ellipsis on outing title
image

* Summary can be up to 3 lines (Chrome), and 300 chars (others), stripping out markdown in a very simple manner.
image
image
(in the last example, content was `Les articles "Vallée d'Aoste ski de randonnée" présentent les [590 itinéraires de ski de randonnée dans la Vallée d’Aoste](/routes#a=280072&act=skitouring) décrits dans le topoguide de Camptocamp, dont [**31 itinéraires de ski de randonnée dans le Valdigne S**](/routes#w=113000,360281,263201,104528,104446,444130,&act=skitouring).`)

### DIFF
--- a/src/components/cards/FeedCard.vue
+++ b/src/components/cards/FeedCard.vue
@@ -13,11 +13,11 @@
 
     <card-row>
       <document-title :document="item.document" class="is-ellipsed has-text-weight-bold" />
-      <span v-if="documentType=='outing'" class="is-nowrap">{{ dates }}</span>
+      <span v-if="documentType=='outing'" class="is-nowrap has-mobile-left-margin">{{ dates }}</span>
     </card-row>
 
     <card-row v-if="locale && locale.summary">
-      <p class="is-ellipsed">{{ locale.summary }}</p>
+      <p class="is-ellipsed is-mobile-max-3-lines-height">{{ locale.summary | stripMarkdown | max300chars }}</p>
     </card-row>
 
     <card-row v-if="images.length!=0">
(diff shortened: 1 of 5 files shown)


cbeauchesne:
It seems to be perfect.

For the markdown, the good way would be to cook summary in API. Do you know if an issue still exists on API ?

brunobesson:
Not sure using cooked markdown is what we need: we want to strip out any HTML content from text (no link, no emphasis, etc).

cbeauchesne:
Stripping out every markdown syntax will be a pain... I quote a good dev on this : 

> A very basic (a.k.a not bullet-proof) filter to remove markdown content from summary

:)

brunobesson:
Not sure I really understand what you mean, appart from the fact that I was pointing out the same kind of default when you handled special chars in respectercestproteger fields 🤣 

In this case, what do you think:? In summary, we generally shouldn't encounter many things appart from links and some emphasis (do you think of other obvious stuff? lists? but then keeping `*` or `-` is not a real problem). In case markdown content is not handled, it will simply be printed out.

cbeauchesne:
> Not sure I really understand what you mean, appart from the fact that I was pointing out the same kind of default when you handled special chars in respectercestproteger fields

Exactly :) 

What I think is that in a perfect world, API feed service must provides locales's cooked data.  And I agree that it's definitly not a big matter.

PR #342, opened 2019-03-21
add dashboard link in the navigation

flob38:
Since we don't know yet what is the best way to integrate the dashboard, but this dashboard deserve a "place" (and a way to access it from the web site), I propose a link in the navigation, near the research box.

### DIFF
--- a/src/views/Navigation.vue
+++ b/src/views/Navigation.vue
@@ -39,6 +39,12 @@
         </div>
       </div>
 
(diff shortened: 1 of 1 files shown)


brunobesson:
This could be a temporary solution, but in the end we have to take a decision at the communty level on how to handle this. IMHO, a kind of switch in the feed could be a nice solution. How responsive is this dashboard (I haven't tested it yet)?

PR #343, opened 2019-03-21
Do not duplicate link

brunobesson:
Use custom content based on property rather than duplicating links

### DIFF
--- a/src/components/generics/links/AddLink.vue
+++ b/src/components/generics/links/AddLink.vue
@@ -2,9 +2,13 @@
   <!-- We alawys display this link, even when user is not logged
     In this case, editions vues will redirect user to /auth -->
   <router-link :to="{name:documentType + '-add', params:{lang:$language.current}, query:query}">
-    <slot>
+    <template v-if="isShortLinkOnMobile">
+      <span class="is-hidden-mobile">{{ $documentUtils.getCreationTitle('outing') | uppercaseFirstLetter }}</span>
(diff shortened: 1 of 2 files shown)


src/components/generics/links/AddLink.vue
```diff
@@ -2,9 +2,13 @@
   <!-- We alawys display this link, even when user is not logged
     In this case, editions vues will redirect user to /auth -->
   <router-link :to="{name:documentType + '-add', params:{lang:$language.current}, query:query}">
-    <slot>
+    <template v-if="isShortLinkOnMobile">
```
cbeauchesne: @brunobesson gosh Oo template inside template ?! what is this feature ?

brunobesson: Well a template is only a way not to output a tag. Can't it be nested?

cbeauchesne: I've learned something today, thank you very much. And actually, I remember that I did not succed to apply this logic, and defaulting on duplicated link with regrets. Good job :)

cbeauchesne: Arf, I was too fast on this one.

The logic of this component is to let devs the choice : 

* use default content : text creationTitle
* or explicitly set the content

It may not depend on mobile/not mobile.

See this for instance : 

image

I will do a Pr with a fix, but I will pay attention to not duplictae the link :)

PR #344, opened 2019-03-21
Use same color for guidebook and forum buttons

brunobesson:
Blues are not different enough => why not keep the same color?


brunobesson:
image

cbeauchesne:
Totally agree

PR #347, opened 2019-03-21
Use noopener or noreferer when applies

brunobesson:


src/components/map/map-layers.js
```diff
@@ -44,7 +44,7 @@ function createIgnSource(title, layer, format = 'jpeg') {
     tileGrid,
     style: 'normal',
     attributions: [
-      '<a href="http://www.geoportail.fr/" target="_blank">' +
+      '<a href="http://www.geoportail.fr/" target="_blank" rel="norefer">' +
```
flob38: norefer -> norefer**er** ?

brunobesson: Right, nice catch!

PR #351, opened 2019-03-21
More details in package.json

cbeauchesne:


package.json
```diff
@@ -1,7 +1,18 @@
 {
-  "name": "vue-camptocamp",
-  "version": "0.1.0",
-  "private": true,
+  "name": "c2c_ui",
+  "description": "https://www.camptocamp.org user interface",
+  "version": "7.0.0beta",
```
brunobesson: Do we really need to start from 7.0?

cbeauchesne: We start from whatever we want. Increasing major version seems to be obvious, but that's a number that may cause some reaction. I've hesitate to skip 7, and go directly to 8 for this reason :) 

Whatever, what is your opinion on this ?

brunobesson:
1.0.0?

cbeauchesne:
As this project is a refactoring from another project, don't you think it's kinda weird to start from 1.0, which means a totally new project ?

@asaunier, you own the legacy of c2c projects, what is your opinion ?

asaunier:
@cbeauchesne I own nothing :P 

Since c2c_ui fully replaces the current UI whose latest release is v6.2.21, it makes sense to refer to it as UI v7.x
What version number will be used on the forums? I'm pretty sure every one will be confused if the new UI is refered to as "v1".

cbeauchesne:
@asaunier 

> What version number will be used when discussing about the new UI on the forums? 

"ui-2019" (a tag on forum), "la maquette", "la nouvelle interface"... Nothing that looks like a version number.

> I'm pretty sure every one will be confused if the new UI is refered to as "v1".

I agree.

brunobesson:
Let it be 7.0.0 then

cbeauchesne:
O, let's go! thanks.

PR #422, opened 2019-04-03
Add documentation for production release

cbeauchesne:


brunobesson:
I'd rather have the documentation in the wiki!,

cbeauchesne:
> I'd rather have the documentation in the wiki!,

I agree, I plan to migrate it all when I'll have some time to do it. BTW, for this one, do you validate the content ?

brunobesson:
> do you validate the content ?

You are missing some pints IMHO:

* `docker-compose.yml` file must be edited to depend on new image
* `docker-compose.yml` must be commited after update (only on first instance, as the other one references the same)
* `http://camptocamp.org:8008/stats` should be mentionned too.

cbeauchesne:
Done. Note for myself, how is it possible that we need to update docker-compose on ui0, and not on ui1 ?

brunobesson:
> we need to update docker-compose on ui0, and not on ui1 ?

This is the same file, there is a symlink.

brunobesson:
P.S. rather than having comments in the bash commands, I would rather have plain text and several bash blocks. But this is not really important

cbeauchesne:
> P.S. rather than having comments in the bash commands, I would rather have plain text and several bash blocks. But this is not really important

My personal experience, as I'm not very familiar with bash comman :  having a comment for each line helps a lot to understand each step, and helps to memorize them :)

PR #440, opened 2019-04-04
Update vue slider dependancy

cbeauchesne:
Major dep update, there is some fix to do... Will close #366


cbeauchesne:
@brunobesson I can't access to snyk report, do you have some info of failure cause ?

brunobesson:
The analysis could not be performed, I don't know why. Let's ignore it.

PR #451, opened 2019-04-04
Add translations for map layers

### LINKED ISSUE #450: Map layers titles are not translated

brunobesson:
Fixes #450

### DIFF
--- a/src/components/map/OlMap.vue
+++ b/src/components/map/OlMap.vue
@@ -17,14 +17,14 @@
               <header v-translate>Base layer</header>
               <div v-for="layer of mapLayers" :key="layer.title" @click="visibleLayer=layer">
                 <input :checked="layer==visibleLayer" type="radio">
-                {{ layer.get('title') }}
+                {{ $gettext(layer.get('title')) }}
               </div>
             </td>
             <td>
               <header v-translate>Slopes</header>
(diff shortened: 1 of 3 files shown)


src/components/map/OlMap.vue
```diff
@@ -17,14 +17,14 @@
               <header v-translate>Base layer</header>
               <div v-for="layer of mapLayers" :key="layer.title" @click="visibleLayer=layer">
                 <input :checked="layer==visibleLayer" type="radio">
-                {{ layer.get('title') }}
+                {{ $gettext(layer.get('title')) }}
```
cbeauchesne: If I'm not wrong, it should be : 

    {{ $gettext(layer.get('title'), 'Map layer') }}

cbeauchesne: BTW, if there is no ambiguity in the english string, you may skip translation context, ie remove second argument.

brunobesson: I have used different context for base layers and slope layers, this is why.
But to answer your first comment: are you sure? Maybe we should add both: one for the extracting tool to add context in the PO file, and one in the effective text so that the tool work properly?

cbeauchesne: Ok, I understand.

For the first comment, it's a feeling. If a context is by far the most common, you can skip it. If they are equivalent in usage, specifying both context is not a bad solution.

Your call :)

PR #564, opened 2019-04-16
change texts in yeti page

jbeilin:
Texts in the Yeti page have been updated : 
- title
- "risque BRA" modified to "danger BRA"
- distance de sécurité to distaces de délestage

Jbeilin


cbeauchesne:
👍 @jbeilin

Issue #570, opened 2019-04-17
Use distinct translations for snow quality and quantity ratings as opposed to condition ratings


brunobesson:
At least in french, gender of tranlation is different (female plural against female singular)

PR #571, opened 2019-04-17
Distinguish translation for snow quality and quantity ratings

brunobesson:
One cannot use the same as condition ratings because (at least in french), gender is different.

A fix for #570

### DIFF
--- a/src/js/constants/Field.js
+++ b/src/js/constants/Field.js
@@ -4,7 +4,7 @@ import utils from '@/js/utils';
 
 const attrs = common.attributes;
 
-// values  can be a string : it describes a common.attributes fields
+// values can be a string : it describes a common.attributes fields
 for (const property of Object.values(fieldsProperties)) {
   if (property.values && typeof property.values === 'string') {
     property.i18nContext = property.values;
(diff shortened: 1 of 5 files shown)


cbeauchesne:
Well, actually, fieldProperties.json is computed with `npm run update-c2c-common` (`tools/update-c2c-common.py` script). It loads data from v6_common repo. So, if we need a modification on this, we must changes v6_common to ensure no divergence between UI and API data model.

I've added a issue in v6_common. I will take this occasion to change format file in V6_common from `py`to `json`, because having a python script in UI is not the best situation.

https://github.com/c2corg/v6_common/issues/64

brunobesson:
Another possibility is to add context (I did that first), but it should be
better this way I think.

cbeauchesne:
> Another possibility is to add context

It's a part of the solution. Context for data from fieldProperties.json is the name of the array. As now (before your PR), context was `condition_ratings`. Now, it's `snow_quality_ratings` and `snow_quantity_ratings`, which is the target for i18n part : https://github.com/c2corg/c2c_ui/pull/571/files#diff-fe7f7ea6b133af95e0b337bcbf5fbfa5R320

Issue #403, opened 2019-04-02
Generic design issues


cbeauchesne:
This issue is a generic container for small and easy-to-fix design issues.

----

* [x] on mobile, margin are too big [Explication](https://forum.camptocamp.org/t/projet-ui-vos-retours-sur-le-design/237297/188?u=charlesb)
* [ ] on mobile, change box orders : tools, licence, association must come after content 
* [ ] hide scrollbar on ImageViewer : [explication](https://forum.camptocamp.org/t/projet-ui-fonctionnalites/234997/282?u=charlesb)
* [x] on history view, language label is in a disabled button. It should be a simple span
* [x] comment button is to wide on small screen and overflow.

brunobesson:
> hide scrollbar on ImageViewer
This has been handled through #454

flob38:
It seems that the 2nd point (hide scrollbar on ImageViewer) is already solved : the scrollbar isn't above the image anymore. Do we hide scrollbar anyway ?

cbeauchesne:
@flob38 , yes, if there is an easy way to remove it (I mean, not a hack). But it's obvioulsy a detail :)

flob38:
There's something strange with the toolbox : associated documents are in the toolbox, so when we put the toolbox after anything else on mobile, the associated documents are after the comments...
![image](https://user-images.githubusercontent.com/24532066/56845940-fc784880-68c8-11e9-810d-a27e6b45b7e7.png)
That's what is required in the list above, but to me, it's make more sense to put associated documents just after associated outings and routes and before comments. What do you think ?

cbeauchesne:
> What do you think ?

It will be a little difficult to totally split data : we are in front of limits of an UI that tries to be as good as possible on mobile AND desktop... 

Another (easy) possibility : add a small arrow that expand/collapse comment box : at least, toolbox (and thus, associated doc) won't be so far.

On an other hand, I think that we should list here what is techniclly possible (eg without falling in a high complexity), and let community choose.

PR #578, opened 2019-04-18
Reduce margin/padding on mobile

flob38:
Related to [issue 403, point 1](https://github.com/c2corg/c2c_ui/issues/403)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -130,6 +130,14 @@
     .page-content{
       margin-left:0;
     }
+
+    .section{
+      padding: 0.75rem !important;
+    }
+
+    .box:not(:last-child), .feed-card{
+      margin-bottom: 0.5rem !important;
+    }
   }
 
   @media screen and (min-width: $tablet) and (max-width: $desktop){
--- a/src/App.vue
+++ b/src/App.vue
@@ -136,7 +136,7 @@
     }
 
     .box:not(:last-child), .feed-card{
-      margin-bottom: 0.5rem !important;
+      margin-bottom: 0.75rem !important;
     }
   }
 


brunobesson:
Could you put screenshots before/after? (I'm lazy, I know)

src/App.vue
```diff
@@ -130,6 +130,14 @@
     .page-content{
       margin-left:0;
     }
+
+    .section{
+      padding: 0.75rem !important;
```
brunobesson: Using `!important` may be a flaw in design. Sorry, I can check this today, but cannot it be avoided?

flob38: Section padding is set in the variables.scss line 62 :
`$section-padding: 1.5rem 1.5rem;`

I don't know how to set it for mobile if it's not with `!important`

flob38:
Screen shots
**Before (Iphone 5)**
image

**After**
image

src/App.vue
brunobesson: Well, media queries do not change the specificity of a selector in css, so both rules, the one defined by bulma and the one you are overriding compete with equals power. In such a case, the last one wins. So the good solution would be to have the custom, app-specific css loaded after the vendor-specific css content.

Alas, there seem to be some problems with the CSS configuration:
* @cbeauchesne states somewhere in the doc that bulma css is included into app chunk instead of vendor-chunk
* if you look at the production site, the webpack plugins seem a little bit messed up, because each css chunk is included twice.

Really, there should not be the need to use `!important`, but the situation is quite a mess... @cbeauchesne  any feedbac on this?

brunobesson:
Thanks for the screenshots.
Is it your choice to have a smaller vertical margin between blocks than the horizontal one? I'd rather have them equal?

flob38:
I reduce vertical margin to save space, but I could put the same.

brunobesson:
Appart from the problem with `!important`, but which surely requires more in-depth work, it looks good to me, thanks!

cbeauchesne:
> Could you put screenshots before/after? (I'm lazy, I know)

Note that I've put in place a full deployment of all branch. To see changes in a branch : 

* https://c2corg.github.io/c2c_ui/master
* https://c2corg.github.io/c2c_ui/design-on-mobile

I'll put on https://c2corg.github.io/c2c_ui/#/ a small page with link to every available builds.

PR #579, opened 2019-04-18
Put toolbox at the end of the page on mobile

flob38:
Related to [issue 403, point 4](https://github.com/c2corg/c2c_ui/issues/403).

Maybe it's not the best way to do it.

### DIFF
--- a/src/views/document/OutingView.vue
+++ b/src/views/document/OutingView.vue
@@ -8,7 +8,7 @@
 
       <div class="column is-3 no-print">
         <map-box :document="document" />
-        <tool-box :document="document" />
+        <tool-box :document="document" class="is-hidden-mobile" />
       </div>
 
       <div class="column is-9 is-12-print">
@@ -107,6 +107,10 @@
 
         <comments-box :document="document" />
 
+      </div>
+
+      <div class="column is-3 no-print is-hidden-tablet">
+        <tool-box :document="document" />
       </div>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


flob38:
I didn't search a lot but can't see a way of doing this.
With v-if, I guess we could avoid duplicate the lines in the render (is-hidden-mobile only put display:none) but not in the code.

By the way, I did this on outing, but there are also the others documents to do...
A global solution would be nicer.
I tried some css stuff like `position: absolute` for the toolbox but it's based on the heigth of the page without the scroll.
image

flob38:
The problem with flexbox is that it's complicated to have 2 vertical items (map and toolbox) beside the main content without nested element (and we don't want to if we want to separate map and toolbox on mobile).
float does that well...

brunobesson:
I understand it may be tricky, maybe impossible to avoid duplicating things. Still, I had to ask 🙂

flob38:
I have no time until tuesday, but I will try something next week.

Le ven. 19 avr. 2019 09:18, Bruno Besson <notifications@github.com> a
écrit :

cbeauchesne:
I'm not a fan of duplication : it's a path that leed us to dam heavy UI. I'd prefer to stick to a conservative way of doing thing, even if the price is sometimes to keep not perfect UI.

I've tried to find a solution last month, but the solution I saw was not the good one : I tried to use css `order` property. But to do this, I must put all box in a flex flow, with direction column. But, the blocking was that it's not possible to control the position of a new column explicitly without knowing the height of each elements.

As now, I do not have any idea of the good solution. Maybe a javascript solution that moves the block ?

flob38:
Duplication is not a good idea. I will try something else more cleaner about the code.

PR #595, opened 2019-04-21
Some minor changes on xreports edition forms #566

cbeauchesne:


brunobesson:
This PR is only part of the modifications projected for xreports?

cbeauchesne:
A part of it, the one which does not requires an API modification

Issue #582, opened 2019-04-18
On mobile, search result lists need a navigation bar on the bottom, or always accessible


brunobesson:
On mobile, the document search filters and navigation link on top do not stay on the screen when scrolling. Thus, one has to scroll all the way back to the top to change page or settings.

Either we ensure that these filters stay visible (which I prefer), or we add another tool bar on the bottom.

![image](https://user-images.githubusercontent.com/2234024/56378058-287e3480-620c-11e9-92a0-d88fd81aea80.png)
![image](https://user-images.githubusercontent.com/2234024/56378074-33d16000-620c-11e9-8d1d-fd2c5cd0413d.png)

PR #599, opened 2019-04-24
DocumentsView - Fixe on top the header and filter sections on mobile

flob38:
Related to issue #582


flob38:
There's something wrong in the code of DocumentsView.vue : it's messy in the syntax highlighting from line 15.
Lint doesn't solve it.
Any idea ?

cbeauchesne:
Be carefull to widely test all use cases : the CSS I wrote look more like a dirty hack than a bullet proof code....

flob38:
Strange, there's no more mess today... I didn't do anything...

PR #601, opened 2019-04-25
Comment button on small screen

flob38:
Related to issue #403

### DIFF
--- a/src/views/document/utils/boxes/CommentsBox.vue
+++ b/src/views/document/utils/boxes/CommentsBox.vue
@@ -187,7 +187,7 @@
   };
 </script>
 
-<style lang="scss">
+<style lang="scss" scoped>
 
   @import '@/assets/sass/variables.scss';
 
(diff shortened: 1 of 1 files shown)


flob38:
I add the scoped specification on css because I don't see why it wasn't there (and make the css for the button work).

cbeauchesne:
> I add the scoped specification on css because I don't see why it wasn't there (and make the css for the button work).

It's because this css target cooked messages from discourse. As vue never know this cooked html, it can't add scoped data style. Have a look on here : 

https://c2corg.github.io/c2c_ui/design-on-mobile4/#/articles/996571/fr/corrections-semi-automatiques-dans-le-topoguide

If you want to use scoped style, you can add a second style tag : 

```
<style lang-"scss">
// style here target anything in the app, 
<s/tyle> 

<style lang-"scss" scoped>
// style here target only html inside my component, 
<s/tyle> 
```

src/views/document/utils/boxes/CommentsBox.vue
```diff
@@ -255,4 +255,9 @@
     }
   }
 
+  .button{
```
cbeauchesne: Put this class inside a new style tag, and remove scoped from the original style tag

PR #624, opened 2019-05-09
Document view - Modify layout columns to layout grid

flob38:
In order to put the toolbox at the end of the page on mobile and small screen, we have studied different options. We must think of a new layout that doesn't put the toolbox in the same column than the other informations that we want to display first on mobile.
Grid layout seems to be the easiest way to display boxes on two columns without nesting elements.

I have not modified the user's profile page because putting the Toolbox at the end of the page is not a good idea (since the user feed has an infinite scroll...).

### What is done : 
We have a basic flex display with boxes that are 100% wide.
For browsers which support the grid layout and if the screen is larger than $tablet, then we display in grid.
Browsers which do not support grid layout (and browsers which do not support @supports in css...) will have the same layout than mobile. => IE (doesn't support @supports) / Opera mini, Blackberry Browser, QQ Browser, Baidu Browser (doesn't support grid layout)

### Pro : 

- It does not duplicate the code
- It remains simple to maintain (anyway we already have a different css on mobile).

### Con : 

- Some browsers that could be used with large screen (IE...) will have the same layout than small screen => 2% of possible users.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -131,9 +131,9 @@
     @media screen and (min-width: $tablet){
       .wrapper{
         display: grid;
-        grid-template-columns: 1fr 2fr;
+        grid-template-columns: 23% auto; //need to fix the first column for the is-ellipsed to work, 1fr 3fr doesn't work
         grid-template-rows: auto 1fr;
-        grid-gap: 0.75rem;
+        grid-gap: 1.5rem;
       }
 
       .item1, .item3{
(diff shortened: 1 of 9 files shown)


cbeauchesne:
@flob38 I close this PR, as the new system will allow us to perform that kind of modification more easily.

flob38:
No problem :)

PR #627, opened 2019-05-12
Improvement of map view design

flob38:
I think the map view wasn't finished : the different fields are displayed without translation and directly on the background.


cbeauchesne:
@flob38, I've merged your PR, because I'll do a release in minutes. Thanks for this fix :)

PR #632, opened 2019-05-14
Small fixes for dashboard's design

flob38:
Add space between avatar and title of topic, in forum-view
Add ellipsis on title, in forum-view


cbeauchesne:
If I remember well, @brunobesson removed the `is-ellipsed` class one month ago.

IMO, we should **not** ellipse text for wide mode (used in feed), we the idea is to use space, and we should ellipse it for dashboard, where text is more dense.

flob38:
Ok, I changed so that only the dashboard is impacted now.

cbeauchesne:
All good to me, feel free to merge.

PR #668, opened 2019-06-10
Move activities icons in associated route links

asaunier:
As for now, when listing routes associated to an outing, the activities icons are shown first:

Capture d’écran 2019-06-10 à 16 30 21

This PR suggests to move them after the route name because:
* the route name is the most important info, thus should be displayed first
* the activities are a route info, like the gradings and heights, it's more consistent to group them
* when several routes are associated, if several activities are available for a given route, it shifts the route names differently and makes it harder to read.

Here is what the change looks like:

Capture d’écran 2019-06-10 à 16 33 28

### DIFF
--- a/src/components/generics/pretty-links/PrettyRouteLink.vue
+++ b/src/components/generics/pretty-links/PrettyRouteLink.vue
@@ -1,11 +1,11 @@
 <template>
   <document-link :document="route" class="pretty-route-link has-hover-background">
+    <document-title :document="route" />
+    <span>&nbsp;</span>
     <activities
       v-if="!hideActivities"
       :activities="route.activities"
       class="is-size-3 has-text-secondary icon-activities" />
-    <span>&nbsp;</span>
-    <document-title :document="route" />,
     <span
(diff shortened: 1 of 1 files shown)


asaunier:
Perhaps it would be more consistent and helpful to add a route icon at the beginning of the line, as for the participants?

cbeauchesne:
I personally prefer icons in first, but it's matter of taste. Just be sure that any changes will be accepted by users, otherwise we'll pay huge cost after :) 

Note that CI have deployed a ready-to-test version of your branch, it's easy to give URL to community to ged some feedback : 

https://c2corg.github.io/c2c_ui/routelink-activities/#/outings/1119866/fr/calanque-d-en-vau-eperon-des-americains
vs
https://c2corg.github.io/c2c_ui/master/#/outings/1119866/fr/calanque-d-en-vau-eperon-des-americains

asaunier:
Sans déconner...

cbeauchesne:
I've forgotten to follow this one? What was the the status, we discuss about it ono the forum ?

PR #669, opened 2019-06-10
Improve image info rendering

asaunier:
This PR makes changes both in the image main pages and in the info section of the images viewer.

## Image main pages

TODO:
* [x] remove filename (who cares?)
* [x] exposure_time: change unit from "ms" to "s"
* [ ] display exposure_time as eg. "1/1600 s", which is the usual way to display it in photography and is much more readable than floats with many digits (eg. 0.00050000002375297)
* [ ] convert datetime to localtime instead of GMT time
* [ ] fnumber: add prefixing "f/" (+ rename title to "Ouverture" instead of "Propriétés de l'image" which is way too general)

Perhaps we could get rid of all those field titles (as in the viewer)?

I don't know yet how to do the uncheck items above. Any suggestions?

BEFORE: 
Capture d’écran 2019-06-10 à 18 12 35

AFTER:
Capture d’écran 2019-06-10 à 18 12 23

## Image viewer info section

TODO:
* [x] remove useless "Settings" / "Infos" section header
* [x] display exposure_time as eg 1/1600 etc.
* [x] limit fnumber to max 1 digit after comma (I have encountered case like "6.30000012345")
* [ ] display datetime as local time instead of GMT time (ditto as for main page, see above)
* [ ] add some some margin after icons
* [ ] add author name

BEFORE:
Capture d’écran 2019-06-10 à 18 24 13

AFTER:
Capture d’écran 2019-06-10 à 18 24 25

### DIFF
--- a/src/components/image-viewer/ImageInfo.vue
+++ b/src/components/image-viewer/ImageInfo.vue
@@ -23,18 +23,14 @@
         <label>{{ $gettext(document.image_type) }}</label>
       </p>
 
-      <h4 class="title is-4 has-text-light" v-translate>
-        Settings
-      </h4>
-
       <ul>
         <li v-if="document.camera_name"> {{ document.camera_name }}</li>
-        <li v-if="document.exposure_time" :title="$gettext('exposure_time')">{{ document.exposure_time }}s</li>
-        <li v-if="document.fnumber" :title="$gettext('fnumber')">f/{{ document.fnumber }}</li>
+        <li v-if="document.exposure_time" :title="$gettext('exposure_time')">1/{{ Math.floor(1/document.exposure_time) }}&nbsp;s</li>
+        <li v-if="document.fnumber" :title="$gettext('fnumber')">f/{{ Math.round(document.fnumber * 10) / 10 }}</li>
         <li v-if="document.focal_length" :title="$gettext('focal_length')">{{ document.focal_length }}&nbsp;mm</li>
-        <li v-if="document.iso_speed" :title="$gettext('iso_speed')">{{ document.iso_speed }} ISO</li>
+        <li v-if="document.iso_speed" :title="$gettext('iso_speed')">{{ document.iso_speed }}&nbsp;ISO</li>
         <li v-if="document.width && document.height" :title="$gettext('resolution')">
-          {{ document.height }} x {{ document.width }} <span translate>pixels</span>
+          {{ document.height }} x {{ document.width }}&nbsp;<span translate>pixels</span>
         </li>
       </ul>
     </div>
@@ -85,5 +81,10 @@
     .image-info{
         background: rgba(0,0,0,0.7);
         padding:1rem;
+    }
+
+    /* FIXME seems not to work */
+    icon-creative-commons, fa-icon {
+        margin-right:1rem;
     }
 </style>
--- a/src/js/constants/fieldsProperties.json
+++ b/src/js/constants/fieldsProperties.json
@@ -373,7 +373,7 @@
         "type": "number",
         "min": 0,
         "max": 0,
-        "unit": "ms"
+        "unit": "s"
     },
     "external_resources":
     {
--- a/src/views/document/ImageView.vue
+++ b/src/views/document/ImageView.vue
@@ -26,7 +26,6 @@
           <field-view :document="document" :field="fields.fnumber" />
           <field-view :document="document" :field="fields.focal_length" />
           <field-view :document="document" :field="fields.iso_speed" />
-          <field-view :document="document" :field="fields.filename" />
           <field-view :document="document" :field="fields.file_size" unit="ko" :divisor="1024" />
           <field-view :document="document" :field="fields.height" />
           <field-view :document="document" :field="fields.width" />


src/components/image-viewer/ImageInfo.vue
```diff
@@ -23,18 +23,14 @@
...
-        <li v-if="document.exposure_time" :title="$gettext('exposure_time')">{{ document.exposure_time }}s</li>
-        <li v-if="document.fnumber" :title="$gettext('fnumber')">f/{{ document.fnumber }}</li>
+        <li v-if="document.exposure_time" :title="$gettext('exposure_time')">1/{{ Math.floor(1/document.exposure_time) }}&nbsp;s</li>
+        <li v-if="document.fnumber" :title="$gettext('fnumber')">f/{{ Math.round(document.fnumber * 10) / 10 }}</li>
         <li v-if="document.focal_length" :title="$gettext('focal_length')">{{ document.focal_length }}&nbsp;mm</li>
-        <li v-if="document.iso_speed" :title="$gettext('iso_speed')">{{ document.iso_speed }} ISO</li>
+        <li v-if="document.iso_speed" :title="$gettext('iso_speed')">{{ document.iso_speed }}&nbsp;ISO</li>
         <li v-if="document.width && document.height" :title="$gettext('resolution')">
-          {{ document.height }} x {{ document.width }} <span translate>pixels</span>
+          {{ document.height }} x {{ document.width }}&nbsp;<span translate>pixels</span>
```
asaunier: What's the point of translating "pixels"?

src/components/image-viewer/ImageInfo.vue
```diff
@@ -86,4 +82,9 @@
         background: rgba(0,0,0,0.7);
         padding:1rem;
     }
+
+    /* FIXME seems not to work */
+    icon-creative-commons, fa-icon {
```
asaunier: Is it the correct way to add styles to elements?

asaunier:
By the way is it really useful to repeat the image title in the viewer info section whereas it is already displayed on top of the viewer?
I would recommend to remove https://github.com/c2corg/c2c_ui/blob/master/src/components/image-viewer/ImageInfo.vue#L13-L16

asaunier:
Concerning 
> convert datetime to localtime instead of GMT time

It would be more readable to show the local time when the image was shot. In my examples above 16:33:06 (instead of 14:33:06+00:00) and 12:32:58 (instead of 10:32:58+00:00) which are much more readable for the visitors to figure the actual time out.

No clue yet how we could do this though. It's not about converting to the current browser time setting but to the local time of the location when the image was shot.
Any suggestion?

src/components/image-viewer/ImageInfo.vue
cbeauchesne: Side not, it's a typo : the good directive is `v-translate`. I guess i've copied the template from V6 without suffisant attention :) 

That said, "pixel" is not universal, I agree with V6 template, it should be translated.

src/components/image-viewer/ImageInfo.vue
cbeauchesne: It's better to add a dedicated class on your template. Do not bother about names collision, the `scoped` attribute on style part make class declared inside this block available only inside the component.

The only thing to take care of is to choose a class that is not used globally, e.g. bulma class, and class declared in [helper.scss](https://github.com/c2corg/c2c_ui/blob/master/src/assets/sass/helpers.scss)

flob38:
> No clue yet how we could do this though. It's not about converting to the current browser time setting but to the local time of the location when the image was shot.
> Any suggestion?

Could you guess the timezone from the country attached to the image ?
image
image
For country which have multiple timezones, it's more complicated.

Modification of API to had the timezone in the area params ?

cbeauchesne:
> Could you guess the timezone from the country attached to the image ?

Actually, the iso format used for image timestamp can contain an UTC offset. I strongly recommend to use this to store the offset (and leave it that way if we do not have the data), over some guessing hack that will be hard to do, and unavoidably buggy.

Thus, we could have a simple preference mechanism that display image's time using user timezone, or image timezone.

flob38:
> the iso format used for image timestamp can contain an UTC offset

I didn't kown that.

> I strongly recommend to use this to store the offset

Do you mean that for existing images this information is lost ?

cbeauchesne:
> Do you mean that for existing images this information is lost ?

Probably.  For them, we could make a bot that guess, but regarding the number of undeterminable cases, I won't recommend that. 

To fix this point properly, here are the steps : 

1. First of all, we have an issue in UI that must be fixed before anything else : we do not read UTC offset on image upload : https://github.com/c2corg/c2c_ui/blob/1ac8d56bd38d9e04901c0ee1fae8f7e73d2f888e/src/components/images-uploader/ImagesUploader.vue#L246 > **the format does not look after seconds !** 😱 
2. Then, add a field in image edition view for this UTC offset value. Its' not a big deal, as it's two simples numeric fields. 
3. Then, we have an unknown on our data : *we don't know if contributors' cameras are mostly well configured**. On smartphone, it's probably ok as clock is synchronized with apple/android servers. But on old-fashion camera, users usually sets the clock hour, without looking of UTC offset. So let's pass some times (one month), and check that the majority of picture taken in France, by french users (english users may have configured an UTC+0) are UTC+2. If, it's pointless to go further, as we can't determine the good value, and we'll implement a smart feature on wrong data... In this case, we should find a way to sensibilize contributor on this.
4. But if majority of data is relevant, this means that majority of camera is well configured, so we can do the very last point : **let the user choose to see hours in his own UTC offset, or picture UTC offset**.

Hope I'm not too obscur ? BTW, even if it looks intimidating, it's not so complicated. it's even a "good first issue", with both technical and data aspect :)

flob38:
> we don't know if contributors' cameras are mostly well configured

That's right. I've just found an recent outing (2019) with images dated in 2005...

asaunier:
> Do you mean that for existing images this information is lost ?

I am not 100% sure but I don't remember those GMT datetime were displayed in v6. I generally pay attention to the pictures datetimes to figure out the time people were at some places and I don't think I had noticed such an issue (times looked consistent I think).

>  we don't know if contributors' cameras are mostly well configured

Definitely :)
I also used to notice silly times quite often. But that wasn't the majority.

flob38:
For exposure time : 
Template : 
`1/{{ exposureTime(document.exposure_time) }} s`
New method : 
```
exposureTime(value) {
        let val = 1/value;
        return Math.round(val);
      }
```
Maybe add a test on exposure_time to see if it's set (and is not zero).

For the fnumber, maybe something with the DecimalFormat, but I can't make it work.
Else, this is working (display f/x with 0.1 accuracy for x) : 
```
fNumber(value){
        let val = value*10;
        return Math.round(val)/10;
      },
```

PR #725, opened 2019-07-31
update dependencies, rollback snyk

cbeauchesne:
Update lot of dependencies, and rollback snyk to 1.198.0


brunobesson:
I couldn't check why snyk is not working anymore. So let's go back, ok. Maybe we could remove it from the build if th fix for lodash is available. I d'ont think there are other issues if you also update vue?

cbeauchesne:
BTW, what is the purpose of snyk, and what is the link with lodash ?

brunobesson:
Snyk is a tool to detect vulnerabilities in your dependencies. Sometimes,
it is not possible to update the version, but a patch can be applied. It is
(was) the case with lodash, but now a new version of the lib exists that
fixes the code. So we could remove the call to the tool in package.json,
and only use the sniffing tool that runs on ci. To be checked, but I assume
it will be safe.

cbeauchesne:
Thanks. Let's keep it, as it seems to work with previous version

Issue #388, opened 2016-12-15
When editing image, impossible to upload new version


mgerbaux:
button "Charger une nouvelle version" make nothing after having selected new file

cbeauchesne:
As it was possible on V5, I consider this issue as a bug.

mgerbaux:
should not be closes, especially for cc-by-sa images (schema)

PR #734, opened 2019-08-10
Upload a new version of an image

cbeauchesne:
Pr for  #388

@mgerbaux you will like this one :) 

@asaunier everything can be done on UI side. But i've one question : does the image backend remove unused image in API DB ? If yes, does it looks in history?

As I can update the document.filename value on images, this script (if it exists) must look on old value of document to determine if an image is used.

### DIFF
--- a/src/components/images-uploader/ImageUploader.vue
+++ b/src/components/images-uploader/ImageUploader.vue
@@ -1,35 +1,34 @@
 <template>
   <div class="card">
-    <div class="card-image img-container" :style="'background-image: url(' + image.src + ')'">
-      <delete-button
-        class="delete-button"
-        @click="$emit('deleteImage', image)" />
+    <div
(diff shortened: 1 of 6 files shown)


asaunier:
> does the image backend remove unused image in API DB ?

I don't think so. When you upload a new version of an image, the former one should be kept in order to be able to view the old versions or even revert to one of them.

At least that's how it worked in v5. I don't remember if it is implemented in the API / images backends.

cbeauchesne:
> everything can be done on UI side

I was wrong 😭 Image backend save my image, but does not process it, neither allow request on it... I must dig into it to understand which event trigger this processing.

cbeauchesne:
Api is modified : if a new filename is sent for a new version of an image, API send publish request to image backend.

Issue #759, opened 2019-08-21
Set main waypoint from first found if not specified


cbeauchesne:
**Community discussion about this feature (mandatory)**
https://forum.camptocamp.org/t/nouvelle-interface-bugs-questions-suggestions/241148/304?u=charlesb

**Is your feature request related to a problem? Please describe.**
When a route is created, main waypoint is often missing. 

**Describe the solution you'd like**
If user set associated waypoints, take the first one, and set it as main waypoint.

**Describe alternatives you've considered**
Alerting the user. But it may be useless friction

**Additional context**
N/A

cbeauchesne:
Mmmmh, should be done on API side... but I'm lazy

PR #760, opened 2019-08-21
When  main waypoint is missing, take the first associated waypoint #759

cbeauchesne:


cbeauchesne:
Can be tested on https://c2corg.github.io/c2c_ui/defaulting-main-waypoint

PR #788, opened 2019-09-24
better documents result width

billyLumberjack:
changed documents-container width depending on displayed document type in order to exploit as max as possible the available space

### DIFF
--- a/src/views/documents/DocumentsView.vue
+++ b/src/views/documents/DocumentsView.vue
@@ -67,7 +67,14 @@
     </div>
 
     <div class="columns result-section" :class="'mobile-mode-' + displayMode">
-      <div v-if="showResults" class="column documents-container" :class="{'is-12': !showMap, 'is-5': documentType == 'route' && showMap,'is-6': documentType != 'route' && showMap}">
+      <div
+        v-if="showResults"
+        class="column documents-container"
+        :class="{
+          'is-12': !showMap,
+          'is-8 is-5-fullhd': showMap && documentType == 'route',
+          'is-8 is-6-fullhd ': showMap && documentType != 'route'
+      }">
 
         <loading-notification :promise="promise" />
 
--- a/src/views/documents/DocumentsView.vue
+++ b/src/views/documents/DocumentsView.vue
@@ -67,7 +67,7 @@
     </div>
 
     <div class="columns result-section" :class="'mobile-mode-' + displayMode">
-      <div v-if="showResults" class="column documents-container" :class="{'is-12': !showMap, 'is-8': showMap}">
+      <div v-if="showResults" class="column documents-container" :class="{'is-12': !showMap, 'is-5': documentType == 'route' && showMap,'is-6': documentType != 'route' && showMap}">
 
         <loading-notification :promise="promise" />
 


cbeauchesne:
Hi @billyLumberjack,

Thank you for your PR :) 

If I understand well, you want to give more place to map when both maps and results are displayed.

* Actually : 8/12 for results, 4/12 for map
* your proposition : 
  * routes  : 5/12 for results, 7/12 for map
  * other documents : half for results, half for map 

Am I right ?

billyLumberjack:
> Hi @billyLumberjack,
> 
> Thank you for your PR :)
> 
> If I understand well, you want to give more place to map when both maps and results are displayed.
> 
> * Actually : 8/12 for results, 4/12 for map
> * your proposition :
>   
>   * routes  : 5/12 for results, 7/12 for map
>   * other documents : half for results, half for map
> 
> Am I right ?

Yes, the map is pretty narrow even on big screens and it is very useful to have a bigger one in my opinion. Furthermore when you have result in list shape lots of space is not used at all.

Thanks for evaluating my PR,
Regards.

SM

cbeauchesne:
Ok, I see.

I've tested on smaller screens, and I think that the left part is really too small (less than 150 px for route cards on my 15 inch laptop). And moving from one third to more than a half is kind of big move :)

But, we can find a solution : Bulma, the CSS framework used, allow to specify class for specific screen width : https://bulma.io/documentation/columns/responsiveness/ and https://bulma.io/documentation/overview/responsiveness/

If I'm not wrong, you target big screens, so `fullhd` (>1408px) in Bulma world. Could you try to use something like `is-8 is-6-fullhhd` or `is-8 is-7-fullhhd` and choose the one that fits the best on Bulma breakpoint ?

billyLumberjack:
Indeed, with small screen actually it does not work so good... I adopted the solution you proposed, it behaves much better. Updates have been already changed
Any kind of feedback is welcome.

Thanks again for your work

SM

cbeauchesne:
IMO, I think that `is-5` or `is-6` is too small for screen around 1400 px. Here is a screen shot of my screen, 1440px width, with `is-6` : 

image

Maybee we should changes the number of cards shown on fullhd ? https://github.com/c2corg/c2c_ui/blob/0c5665045b608c2a22fa20229b354cfbc9c4a7c4/src/views/documents/DocumentsView.vue#L88-L89

That said, I'm not the one who decide : all decisions are made by the community of contributors. Unfortunatly, discussions are mainly made in french. I will propose this change, and translate feedbacks here. Are you also billyLumberjack on camptocamp forum ?

cbeauchesne:
-> https://forum.camptocamp.org/t/affichage-des-resultats-de-recherche/250817

billyLumberjack:
Wow ! what an efficiency, thanks @cbeauchesne for moving my proposal to the forum.
I work in Nice but my french is not the best, I will do my best to follow up the discussion and give my contribution.

Keep in touch

cbeauchesne:
Ok, somme feedbacks: 

* Globally ok for the change
* Does you have seen that you can put the map in full width (hiding table/cards) ? 
* Does it make a sense to have a different width for table/cards : two peoples remarks that when you enable tables, the result is not so good

My personal two cents : 

* Keep it simple principe  : Do not use different width for routes and other document. I'd use 50%
* on cards mode, display only two cards per row (i/o 3), the result is really too small 1400px width (see first screenshot on forum)
* Have a try to use 7/12 on table mode, see if it's better.

cbeauchesne:
Hi @billyLumberjack,

I plan to make a release next week, and as I've exposed your proposal on the forum, I'd like to include this change in this release.

Do you mind if I make this commit myself ?

PR #818, opened 2019-10-17
Translations

cbeauchesne:


cbeauchesne:
On purpose, I've splited this in 3 commits, if anyone wants to check each step.

PR #822, opened 2019-10-21
update dependancies

cbeauchesne:


cbeauchesne:
yes, ol, ag-grid, and even vue-router needs more attention. IMO, it's ok to skip most of updates as we do not use edge features.

PR #843, opened 2019-11-08
YETI: add links in UI

iamvdo:
This PR adds YETI links inside UI:

- on `/routes` page (at the top)
- on each `/routes/{id}` pages when activity is one of `skitouring`, `snow_ice_mixed`, `ice_climbing` or `snowshoeing` (below map, same line as GPX, KML buttons)
    - document data will be loaded inside YETI, on top of YETI layer
    - a go-back link to that route is also added

All changes are made inside YETI page and/or components.
Only one is made inside `OLMap` component: I added ability to force `minZoomLevel` when map is fitted to one document (allow YETI to set minimum accepted zoom for computation, when a document is loaded)

### DIFF
--- a/src/components/yeti/YetiButton.vue
+++ b/src/components/yeti/YetiButton.vue
@@ -19,7 +19,7 @@
         return this.$route.name;
       },
       href() {
-        return '/yeti' + (this.document ? '/' + this.document.document_id : '');
+        return { name: 'yeti', params: { 'document_id': this.document ? this.document.document_id : null } };
       },
       isRouteDocumentType() {
         return this.documentType === 'route' || this.documentType === 'routes';
--- a/src/components/yeti/YetiButton.vue
+++ b/src/components/yeti/YetiButton.vue
@@ -1,0 +1,36 @@
+<template>
+  <router-link v-if="showButton" :to="href" class="button">
+    <icon-yeti class="icon" />
+    <span v-if="document">Voir dans YETI</span>
+    <span v-else>YETI</span>
+  </router-link>
+</template>
+
+<script>
+  export default {
+    props: {
+      document: {
+        type: Object,
+        default: null
+      }
+    },
+    computed: {
+      documentType() {
+        return this.$route.name;
+      },
+      href() {
+        return '/yeti' + (this.document ? '/' + this.document.document_id : '');
+      },
+      isRouteDocumentType() {
+        return this.documentType === 'route' || this.documentType === 'routes';
+      },
+      isActivitiesDocument() {
+        const activities = ['skitouring', 'snow_ice_mixed', 'ice_climbing', 'snowshoeing'];
+        return this.document.activities.some(activity => activities.includes(activity));
+      },
+      showButton() {
+        return this.isRouteDocumentType && (this.document ? this.isActivitiesDocument : true);
+      }
+    }
+  };
+</script>
--- a/src/js/vue-plugins/font-awesome-config.js
+++ b/src/js/vue-plugins/font-awesome-config.js
@@ -34,6 +34,7 @@ import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons/faExclama
 import { faExpand } from '@fortawesome/free-solid-svg-icons/faExpand';
 import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons/faExternalLinkAlt';
 import { faEye } from '@fortawesome/free-solid-svg-icons/faEye';
+import { faExclamation } from '@fortawesome/free-solid-svg-icons/faExclamation';
 import { faFilter } from '@fortawesome/free-solid-svg-icons/faFilter';
 import { faFlag } from '@fortawesome/free-solid-svg-icons/faFlag';
 import { faFlagCheckered } from '@fortawesome/free-solid-svg-icons/faFlagCheckered';
@@ -72,6 +73,7 @@ import { faSignInAlt } from '@fortawesome/free-solid-svg-icons/faSignInAlt';
 import { faSignOutAlt } from '@fortawesome/free-solid-svg-icons/faSignOutAlt';
 import { faSortAmountUp } from '@fortawesome/free-solid-svg-icons/faSortAmountUp';
 import { faStar } from '@fortawesome/free-solid-svg-icons/faStar';
+import { faSquare } from '@fortawesome/free-solid-svg-icons/faSquare';
 import { faSun } from '@fortawesome/free-solid-svg-icons/faSun';
 import { faTachometerAlt } from '@fortawesome/free-solid-svg-icons/faTachometerAlt';
 import { faTag } from '@fortawesome/free-solid-svg-icons/faTag';
@@ -240,6 +242,7 @@ export default function install(Vue) {
     faExpand,
     faExternalLinkAlt,
     faEye,
+    faExclamation,
     faFilter,
     faFlag,
     faFlagCheckered,
@@ -278,7 +281,7 @@ export default function install(Vue) {
     faSignOutAlt,
     faSortAmountUp,
     faStar,
-    //    faSquare,
+    faSquare,
     faSun,
     faTachometerAlt,
     faTag,
--- a/src/js/vue-plugins/font-awesome-config.js
+++ b/src/js/vue-plugins/font-awesome-config.js
@@ -30,11 +30,11 @@ import { faCompass } from '@fortawesome/free-solid-svg-icons/faCompass';
 import { faCompress } from '@fortawesome/free-solid-svg-icons/faCompress';
 import { faDatabase } from '@fortawesome/free-solid-svg-icons/faDatabase';
 import { faEdit } from '@fortawesome/free-solid-svg-icons/faEdit';
+import { faExclamation } from '@fortawesome/free-solid-svg-icons/faExclamation';
 import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons/faExclamationCircle';
 import { faExpand } from '@fortawesome/free-solid-svg-icons/faExpand';
 import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons/faExternalLinkAlt';
 import { faEye } from '@fortawesome/free-solid-svg-icons/faEye';
-import { faExclamation } from '@fortawesome/free-solid-svg-icons/faExclamation';
 import { faFilter } from '@fortawesome/free-solid-svg-icons/faFilter';
 import { faFlag } from '@fortawesome/free-solid-svg-icons/faFlag';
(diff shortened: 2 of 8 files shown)


src/components/yeti/YetiButton.vue
```diff
@@ -0,0 +1,36 @@
...
+        type: Object,
+        default: null
+      }
+    },
+    computed: {
+      documentType() {
+        return this.$route.name;
+      },
+      href() {
+        return '/yeti' + (this.document ? '/' + this.document.document_id : '');
```
cbeauchesne: It's better to use the model object for urls : 

```suggestion
        return {name : 'yeti', params: {'document' : this.document ? this.document.document_id : null }};
```

src/components/yeti/YetiButton.vue
```diff
@@ -0,0 +1,36 @@
...
+      },
+      isRouteDocumentType() {
+        return this.documentType === 'route' || this.documentType === 'routes';
```
cbeauchesne: Avoid property that can fail. And document ships a property type : 
 
```suggestion
        return this.document ? this.document.type === 'r' : False;
```

src/components/yeti/YetiButton.vue
```diff
@@ -0,0 +1,36 @@
...
+    props: {
+      document: {
+        type: Object,
+        default: null
+      }
+    },
+    computed: {
+      documentType() {
+        return this.$route.name;
+      },
```
cbeauchesne: With proposition for isRouteDocumentType, this property no more needed

src/components/yeti/YetiButton.vue
```diff
@@ -0,0 +1,36 @@
...
+      href() {
+        return '/yeti' + (this.document ? '/' + this.document.document_id : '');
+      },
+      isRouteDocumentType() {
+        return this.documentType === 'route' || this.documentType === 'routes';
+      },
+      isActivitiesDocument() {
+        const activities = ['skitouring', 'snow_ice_mixed', 'ice_climbing', 'snowshoeing'];
```
cbeauchesne: Avoir property that can fail 

```suggestion
        if (!this.document) {
            return False;
        }
        const activities = ['skitouring', 'snow_ice_mixed', 'ice_climbing', 'snowshoeing'];
```

src/components/yeti/YetiButton.vue
```diff
@@ -0,0 +1,36 @@
...
+        return this.document.activities.some(activity => activities.includes(activity));
+      },
+      showButton() {
+        return this.isRouteDocumentType && (this.document ? this.isActivitiesDocument : true);
```
cbeauchesne: If isActivitiesDocument cannot fails, this could be simplier : 

```suggestion
        return this.isRouteDocumentType && this.isActivitiesDocument;
```

src/js/vue-plugins/font-awesome-config.js
```diff
@@ -34,6 +34,7 @@ import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons/faExclama
 import { faExpand } from '@fortawesome/free-solid-svg-icons/faExpand';
 import { faExternalLinkAlt } from '@fortawesome/free-solid-svg-icons/faExternalLinkAlt';
 import { faEye } from '@fortawesome/free-solid-svg-icons/faEye';
+import { faExclamation } from '@fortawesome/free-solid-svg-icons/faExclamation';
```
cbeauchesne: Can you keep the alphabetical order please ? for maintenance, it helps a lot.

src/js/vue-plugins/font-awesome-config.js
```diff
@@ -72,6 +73,7 @@ import { faSignInAlt } from '@fortawesome/free-solid-svg-icons/faSignInAlt';
 import { faSignOutAlt } from '@fortawesome/free-solid-svg-icons/faSignOutAlt';
 import { faSortAmountUp } from '@fortawesome/free-solid-svg-icons/faSortAmountUp';
 import { faStar } from '@fortawesome/free-solid-svg-icons/faStar';
+import { faSquare } from '@fortawesome/free-solid-svg-icons/faSquare';
```
cbeauchesne: Can you keep the alphabetical order please ? for maintenance, it helps a lot.

src/js/vue-plugins/font-awesome-config.js
```diff
@@ -240,6 +242,7 @@ export default function install(Vue) {
     faExpand,
     faExternalLinkAlt,
     faEye,
+    faExclamation,
```
cbeauchesne: Can you keep the alphabetical order please ? for maintenance, it helps a lot.

src/js/vue-plugins/font-awesome-config.js
```diff
@@ -278,7 +281,7 @@ export default function install(Vue) {
     faSignOutAlt,
     faSortAmountUp,
     faStar,
-    //    faSquare,
+    faSquare,
```
cbeauchesne: Can you keep the alphabetical order please ? for maintenance, it helps a lot.

src/js/vue-plugins/router.js
```diff
@@ -62,7 +62,7 @@ const routes = [
   { path: '/following', name: 'following', component: FollowingView },
   { path: '/preferences', name: 'preferences', component: PreferencesView },
   { path: '/mailinglists', name: 'mailinglists', component: MailingListsView },
-  { path: '/yeti/:page?', name: 'yeti', component: YetiView },
+  { path: '/yeti/:document(\\d+)?/:page?', name: 'yeti', component: YetiView },
```
cbeauchesne: Stick to the document_id convention

```suggestion
  { path: '/yeti/:document_id(\\d+)?/:page?', name: 'yeti', component: YetiView },
```

src/views/document/utils/boxes/MapBox.vue
```diff
@@ -21,15 +21,22 @@
...
-    <div
-      v-if="document.geometry && (document.geometry.geom_detail || documentType == 'waypoint')"
-      class="buttons is-centered">
-      <button class="button is-primary is-small" @click="downloadGpx">
-        GPX
-      </button>
-      <button class="button is-primary is-small" @click="downloadKml">
-        KML
-      </button>
+    <div class="columns is-multiline is-mobile is-clearfix">
```
cbeauchesne: Can you provide a screenshot of this? When we change UI, we must be discussed on forum (otherwise, we face big difficulties :) )

src/views/portals/YetiView.vue
```diff
@@ -504,7 +515,9 @@
         mountains: {},
         visibleMountains: {},
         promiseMountains: null,
-        showMountainsList: false
+        showMountainsList: false,
+
+        documents: null
```
cbeauchesne: I see that you use `documents` only for map view, and several times as `documents[0]`. Would it be preferable to have `document`property, and call `map-view` with `[document]` ?

src/views/portals/YetiView.vue
```diff
@@ -566,6 +579,21 @@
 
     mounted() {
       this.check();
+      const doc = this.$route.params.document;
+      const lang = this.$language.current;
+      if (doc) {
+        c2c['route'].getCooked(doc, lang).then(doc => {
```
cbeauchesne: Can you have a look on other place where we load data ? Basically, we use something like : 

```
This.promise = load() ..

properties {
  data() {
    return this.promise.data
}
``` 

And you have a `<loading notification> component` that display errors.

iamvdo:
@cbeauchesne Thanks for your suggestions. 

For the first point, I'm trying to add link for `/routes` page AND `routes/id` pages where activity is one mentioned. So i'm testing `this.$route.name` cause there are no document on `/routes` page, so I cannot rely on `document.type`?

iamvdo:
Pour l'ajout du bouton sous la carte (avant/après)

yeti-before-after

cbeauchesne:
> @cbeauchesne Thanks for your suggestions.
> 
> For the first point, I'm trying to add link for `/routes` page AND `routes/id` pages where activity is one mentioned. So i'm testing `this.$route.name` cause there are no document on `/routes` page, so I cannot rely on `document.type`?

Oh, I understand. I think you use a bad pattern. Enterily hiding a component based on outside logic is weird. If you include the component, the component itself should display something, whatever the URL. So, I would do this : 

* on DocumentsView, `<yeti-button v-if="documentType==='route'" />
* remove the `v-if` on the root element of yeti-button
* And then, inside this component, simply use double behavior : 
  * if document is provided, links to `/yeti/<document.document_id>` (and simply raise an excpetion if `document.type` is not `r`
  * else, links to `/yeti/<document.document_id>`

cbeauchesne:
> * on DocumentsView, `<yeti-button v-if="documentType==='route'" />`

Erratum : Not in `DocumentsView`, it's in `QueryItems`

iamvdo:
> Oh, I understand. I think you use a bad pattern

You're right. I'll wait for discussion on forum before updating the code, but I can't find it. Could you share the link?

Note that I added YETI buttons there, based on discussion with @jbeilin and @brunobesson.

cbeauchesne:
>  but I can't find it. Could you share the link?

Unfortunately, it's on the private part of the forum (only accessible to members of c2c association). I'm actively working on changing this, and make things a little bit more open, but it takes time...

Really sorry about the lack of clearness on the spec, I'll get you updated here as long as the discussion is not open.

As now, 

* all good for the link in `routes/document_id` page 
* for the `/routes` page, two users against (including myself :) ) : lack of place, and it's not relevant if user does not look for routes concerned by YETI
* two users (including myself) for having a link in `/topoguide` page (so 1 click far from any place in UI, as it's in the menu).

jbeilin:
> As now,
> 
>     * all good for the link in `routes/document_id` page

Good ! 
> 
>     * for the `/routes` page, two users against (including myself :) ) : lack of place, and it's not relevant if user does not look for routes concerned by YETI
> 
>     * two users (including myself) for having a link in `/topoguide` page (so 1 click far from any place in UI, as it's in the menu).

I am ok to put a link in a seventh circle on the topoguide page (https://www.camptocamp.org/topoguide) as shown on this image but the link on the right of the buttons in the itineraries page was proposed by C2C. 

I think I prefer the first solution (circle in topoguide page) but we will accept the other one if you prefer it. We just have to think that the december, 3rd is very soon !

lien_yeti

BTW, as th pull request has been accepted, when the new version oh the yeti page will be visible ?

cbeauchesne:
@iamvdo : https://forum.camptocamp.org/t/implementation-de-la-version-officielle-de-yeti/250654/31

cbeauchesne:
@iamvdo @jbeilin go for the link on `/topoguide` page

iamvdo:
Closed. See #867

Issue #852, opened 2019-11-11
Change translation icon


cbeauchesne:
Use something like this : 

![image](https://user-images.githubusercontent.com/11915659/68613965-33f2bf00-04c0-11ea-9f02-c818f2228972.png)

Discussion : https://forum.camptocamp.org/t/icones-dans-les-documents/251535/18?u=charlesb

PR #855, opened 2019-11-11
[WIP] Change translatin icon 

cbeauchesne:
Issue #852


cbeauchesne:
finally using a simple globe, only this one is readable on this size

PR #871, opened 2019-11-16
Use Dayjs instead of moment

brunobesson:
API is almost identical, gain is 2 to 4 percents depending on how you compare sizes.

I chose to keep moment name, but there is no problem in changing it, I have no strong opinion on it.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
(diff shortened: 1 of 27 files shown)


cbeauchesne:
I was damn lazy when I choose this name, so IMO, we should remane it to avoid some legacy in naming that will introduce some blur for futur dev.

But do not do the same mistake, use something explicit like `dateUtils` or `dateTools`

PR #876, opened 2019-11-18
YETI Hotfix: Map can't be moved and/or zoomed when a document is loaded 

iamvdo:
Fix: Pass document as array from computed properties. Otherwise, a new array is passed every time the map move (and so, is fitted again)


cbeauchesne:
@iamvdo I trust you on this (no time to check...) can I merge ?

iamvdo:
@cbeauchesne Yep :)

Issue #887, opened 2019-11-21
Bug: Zoom events on OLMap are fired only when zoom is integer


iamvdo:
On YETI, when a user zooms using touch (on mobile) and/or trackpad (when pinch, scroll), zoom events are not fired, as they are no longer integer. So, computation could no longer be done (as zoom will never be updated).

I will fix it, but I need your point of view on a solution:

- Fire all zoom events from `OLMap.vue`
- or, (for me a better solution) remove `change:resolution` event and no longer fire zoom events from `OLMap` (seems that only YETI uses them, added in 605e6da8698f), and let YETI do its own work (already a `moveend` event that can be used)

brunobesson:
This is a bug I found some time ago, I thought I had reported it? Nevermind.

Do you mean that using the trackpad doesn't end with an integer zoom?

iamvdo:
> Do you mean that using the trackpad doesn't end with an integer zoom?

Yes. As far as you started to pinch and/or scroll, zoom is decimal

brunobesson:
Then maybe there is a way to force it to end with an integer value?

brunobesson:
When I zoom in/out using my touch pad, the zoom info testing if one zoomed enough to compute data still displays integer values?

![image](https://user-images.githubusercontent.com/2234024/69331299-cc253c80-0c54-11ea-89a3-900747b2fb63.png)

iamvdo:
>Then maybe there is a way to force it to end with an integer value?

Maybe. But on small screens, it may be a good option to have mid-level zoom to better cover a specific area

>When I zoom in/out using my touch pad, the zoom info testing if one zoomed enough to compute data still displays integer values?

Right now, yes. Because map only fires those zoom events. When the bug will be fixed, we will show decimal zooms (floored to one decimal), eg. 12.9

iamvdo:
Oh, I recall your issue, it was https://github.com/c2corg/c2c_ui/pull/780#issuecomment-534241906

Did not understand at that time. Sorry

iamvdo:
@cbeauchesne Any advice?

PR #895, opened 2019-11-27
Yeti updates

iamvdo:
Last bunch of updates before next week.
Mostly small improvements, bug fixes and wording.

Thanks for review!

### DIFF
--- a/src/components/map/OlMap.vue
+++ b/src/components/map/OlMap.vue
@@ -379,8 +379,6 @@
         // keyboardEventTarget: this.keyboardEventTarget,
       });
 
-      this.view.on('change:resolution', this.onChangeResolution);
-
       this.map.on('moveend', this.sendBoundsToUrl);
       this.map.on('moveend', this.getBiodivSportsAreas);
       if (this.protectionAreasVisible) {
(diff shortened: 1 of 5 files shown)


src/components/map/OlMap.vue
```diff
@@ -702,15 +700,6 @@
         return extent;
       },
 
-      onChangeResolution(event) {
```
cbeauchesne: Why removing this function ? I think it's useful for filtering on `/routes`, `/outings`...

src/js/libs/ol.js
```diff
@@ -77,6 +77,7 @@ export default {
   },
 
   geom: {
+    LineString,
```
cbeauchesne: I can't see where this module is used ?

iamvdo: Here: 

https://github.com/c2corg/c2c_ui/blob/283c7f575de29a426446e31f83ee31a70f85f2df/src/views/portals/YetiView.vue#L952

Basically, it draws linestrings instead of polygon, to improve performance. See https://github.com/openlayers/openlayers/issues/10139

I notice that @ahocevar just replied to use `ol/layer/VectorImage` instead, but it's only v6. I will add `renderMode: 'image'` to increase perf (even if linestrings are still needed)

src/components/map/OlMap.vue
iamvdo: As said in #887, this function was added in first YETI PR. It emits a `zoom` event, and no `OLMap` uses it, apart the one inside YETI.

But I can revert it if you want?

cbeauchesne: No perfect, I believed it was used somewhere else, my bad.

Issue #884, opened 2019-11-20
Feature: Remove entry "My mailing lists" in the user menu


asaunier:
Avalanche mailing lists are no longer available, it is no use giving access to this page.

asaunier:
Perhaps only those lines to remove
https://github.com/c2corg/c2c_ui/blob/f7bf57f5e0857b185447b64c6db505d20dec8a47/src/views/Navigation.vue#L192-L196
?

asaunier:
Someone is wondering why he does not receive the nivo bulletins by email whereas he has subscribed in the (old inactive) ML form page:
https://forum.camptocamp.org/t/bulletins-nivo/253620/2

Not sure the nivo ML will come back one day, at least not at short term => it is better to at least remove the menu item. Perhaps we should also remove the UI page and API service (but not in this issue).

PR #903, opened 2019-12-05
Remove mailing-lists navigation menu item

asaunier:
Related to https://github.com/c2corg/c2c_ui/issues/884

NOT TESTED!


cbeauchesne:
If we remove this page, better to remove also the view and the route : 

https://github.com/c2corg/c2c_ui/search?q=mailinglists&unscoped_q=mailinglists

asaunier:
> If we remove this page, better to remove also the view and the route

Right.
But I don't have time to do it thus won't.

cbeauchesne:
> If we remove this page, better to remove also the view and the route

Done

PR #910, opened 2019-12-07
Feature/filter whatsnew

momomaniac:
update of recent documents interface to go with API change allowing filtering on doc type, completeness and license type

### DIFF
--- a/src/views/wiki/WhatsNewView.vue
+++ b/src/views/wiki/WhatsNewView.vue
@@ -6,20 +6,20 @@
     <h3> Filter recent documents </h3>
     <span class="feed-filters">
 
-    <query-item
-      :key="type_field.name"
-      :field="type_field"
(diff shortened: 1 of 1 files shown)


momomaniac:
the UI form does not support all functions of the API interface:
API allows for including and excluding filters (by preceding with -)
UI interface can only use including (restriction) filters.
However, filters typed manually in the URL will be handed over to the API

src/views/wiki/WhatsNewView.vue
```diff
@@ -88,10 +113,36 @@
     },
 
     computed: {
+
       loading() {
         return this.promise ? this.promise.loading : false;
       },
 
+      type_field() {
```
brunobesson: use camelcase, please :)

cbeauchesne:
Thank's for the head's up bot :)

PR #911, opened 2019-12-07
sorting documents

momomaniac:
…rently only supported in list mode)

TODO:
- read sort state from URL
- handle multi-sort
- fix sorting for some columns (ex: quality does not accept sorting)


momomaniac:
the feature is now working along with the API mod https://github.com/c2corg/v6_api/pull/758
example:
image
- clickable feature is only active in _list view_, but selection remains active in tile view via the URL
- multiple sort can be used by shift-clicking several column headers.

momomaniac:
current limitations:
sorting does not work on __some fields__ yet
- for example the quality field. I tried a workaround in https://github.com/c2corg/c2c_ui/pull/911/commits/09588ddd9462d2926cd30255491bcf92daa8989e, but this may lead to side impacts
- title cannot be sorted because not indexed in elasticsearch
- the custom cell renderers (quality, activity, etc.) do not seem to work well with sorting => custom sort criteria might be used

there is need for some testing with different columns and some discussion on the following questions:
- do all the columns work
- do all the columns work as expected
- is ascending/descending logical
- shall there be a better transparent error handling for criteria that cannot be sorted (currently the user clicks and get an error, or nothing is happening)
- user interface for tile view ???

brunobesson:
I do not think there is a need to allow sorting on all fields.
For example, I don't see the need to sort on title.

For such fields, sorting could be simply disabled.

src/components/datatable/DocumentsTable.vue
```diff
@@ -95,12 +98,52 @@
...
+          this.gridApi.setSortModel(sort_model);
+        }
       },
 
       onHover(event) {
         this.$emit('highlightDocument', event.data);
       },
 
+      sortChanged(event) {
+        // console.log(event);
```
brunobesson: don't leave this comments in the PR

src/components/datatable/DocumentsTable.vue
```diff
@@ -95,12 +98,52 @@
...
+        // console.log(this.gridOptions.columnApi.getColumnState());
+        // console.log(this.gridOptions.api.getSortModel());
+        const sortM = this.gridApi.getSortModel();
+        const query = Object.assign({}, this.$route.query);
+        let sort_list = [];
```
brunobesson: please use camelCase for variables, for consistency (not only here)

src/components/datatable/DocumentsTable.vue
```diff
@@ -95,12 +98,52 @@
     methods: {
       onGridReady(params) {
         this.gridApi = params.api;
+        let sort_model = [];
+        const sort_URL = this.$route.query['sort'];
+        if (sort_URL) {
+          let sort_item = {};
```
brunobesson: there is no check on validity of the URL. Could a regexp be used, e.g. ?

momomaniac:
OK, implemented some comments.
still open: URL validation and disabling some search criteria, this should be somehow synchronized with the index definitions in c2c_common

src/components/datatable/DocumentsTable.vue
momomaniac: I had a look at this, the URL is already slightly error-tolerant, the code tries to parse the URL and rewrite the sort model. If a column name is not spelled correctly, the sort instruction is dropped and the URL updated.
What would you expect to do after validity check ? User-feedback ? Isn't it better to have a silent robust mechanism. Or are you afraid of a vulnerability ? The query is parsed twice afterwards (split tokens, configure sort model) with no active string, so I don't see a way to produce a dangerous behaviour, but I am very bad in security topics, so I may overlook something.

momomaniac:
I cleaned a bit the feature, it is now ready for testing. @cbeauchesne, can we deploy a demo ?

It only works in the table view, but the sort options are kept when switching to thumbnail view.

Columns that should not be sortable are deactivated, but I am still not sure about correct sorting behavior in all cases. For example xreport `event_type` does not seem to sort well in the API, but maybe things improve when updating the SERAC DB as there will not be multiple fields any more.

Test instructions:
- use API with [PR 758](https://github.com/c2corg/v6_api/commit/bc1417ca04c50666a7ef164cb156e7d78e9e5d18)
- open a document view (outings, routes, waypoints, xreports, etc.)
- switch to table/grid view
- click on any column header with double arrow (will switch to single arrow after click) non-sortable columns are not clickable
- hold shift and click second column header for multi-sort
- test with offset/pages, filters, etc.

cbeauchesne:
@momomaniac There is a mechanism that publish all our PR on GH-pages. Unfortunately, I made a misstake when I code this : I forgot to handle wheen you have a slash in branch name :) 

I'm fixing this on this branch.

cbeauchesne:
https://c2corg.github.io/c2c_ui/ and https://c2corg.github.io/c2c_ui/feature-sort_documents/#/ :)

cbeauchesne:
@momomaniac Here is the feedback thread : https://forum.camptocamp.org/t/tri-besoin-de-testeurs/254178

src/components/datatable/DocumentsTable.vue
```diff
@@ -46,6 +48,25 @@
     return Object.assign(result, options);
   }
 
+  // comparator functions for custom sort of "complicated" columns
+  // these functions could be removed if the sort function of
+  // AG grid could be deactivated since documents are expected to
+  // be given in correct order by the backend
+  function condComp(cond1, cond2) {
+    const conds = { 'excellent': 0, 'good': 1, 'average': 2, 'poor': 3, 'awful': 4 };
```
cbeauchesne: You can use array defined in `/src/js/constants/common.json` as a source for your map. It does not duplicate this list and is more maintenable.

Same for other comparator

src/components/datatable/DocumentsTable.vue
```diff
@@ -46,6 +48,25 @@
     return Object.assign(result, options);
   }
 
+  // comparator functions for custom sort of "complicated" columns
+  // these functions could be removed if the sort function of
+  // AG grid could be deactivated since documents are expected to
+  // be given in correct order by the backend
+  function condComp(cond1, cond2) {
```
cbeauchesne: Please use as possible explicit names (and blame me when I didn't do this :) ) 

```suggestion
  function conditionComparator(cond1, cond2) {
```

src/components/datatable/DocumentsTable.vue
```diff
@@ -46,6 +48,25 @@
...
+    const conds = { 'excellent': 0, 'good': 1, 'average': 2, 'poor': 3, 'awful': 4 };
+    return conds[cond1] - conds[cond2];
+  }
+
+  function qualComp(qual1, qual2) {
```
cbeauchesne:
```suggestion
  function qualityComparator(qual1, qual2) {
```

src/components/datatable/DocumentsTable.vue
```diff
@@ -46,6 +48,25 @@
...
+    const quals = { 'empty': 0, 'draft': 1, 'medium': 2, 'fine': 3, 'great': 4 };
+    return quals[qual1] - quals[qual2];
+  }
+
+  function avComp(av1, av2, n1, n2, isInverted) {
```
cbeauchesne:
```suggestion
  function avalancheLevelComparatorComp(av1, av2, n1, n2, isInverted) {
```

src/components/datatable/DocumentsTable.vue
```diff
@@ -46,6 +48,25 @@
...
+    const avs = { 'level_ns': 0, 'level_1': 1, 'level_2': 2, 'level_3': 3, 'level_4': 4, 'level_5': 5 };
```
cbeauchesne: Is it `level_ns` or `level_na` ? (if first, good argument for using array for common.js)

src/components/datatable/DocumentsTable.vue
```diff
@@ -66,7 +87,8 @@
 
     data() {
       return {
-        columnDefs: null
+        columnDefs: null,
+        dColumnDefs: { sortable: true, unSortIcon: true }
```
cbeauchesne: `dColumnDefs` stands for ? Maybe use a more explicit name ?

src/components/datatable/DocumentsTable.vue
```diff
@@ -123,47 +187,47 @@
...
+            getColDef(this, fields.quality, { cellRendererFramework: MarkerQuality, width: 30, comparator: qualComp })
           ];
         }
 
         if (this.documentType === 'book') {
           this.columnDefs = [
             getColDef(this, fields.title, { cellRendererFramework: DocumentLink }),
             getColDef(this, fields.book_types),
-            getColDef(this, fields.author),
+            getColDef(this, fields.author, { sortable: false }),
```
cbeauchesne: I think it would be better to modify `/src/js/constants/fieldsProperties.json` and add a property `sortable`. As most of fields are sortable (is it true ? if no, inversed logic), if it's not defined, it means it's sortable. If it's defined and `True`, it's sortable. Then `getColDef()` will look like something : 

```js

  function getColDef(vm, field, options) {
    options = options || {};
    const result = {
      headerName: vm.$gettext(field.name),
      field: field.name,
      _fieldDefinition: field,
      cellRendererFramework: options.cellRendererFramework || DocumentField,
      resizable: true,
      sortable: field.sortable === undefined || field.sortable === True
      // if most of fields are NOT sortable, then undefined means False: 
      // sortable: field.sortable === True
    };
    return Object.assign(result, options);
  }
```

cbeauchesne:
@b_b, cna you grant me access to snyk? There is something that block this PR, and I can't see why

brunobesson:
@cbeauchesne just rebase the branch on master, it should be enough.

cbeauchesne:
@b_b I do not have access to it :(

brunobesson:
@cbeauchesne we need to sort this out. Not much time for this right now though :(

cbeauchesne:
thank's for the head's up stale bot. 

@momomaniac what was the status here ?

brunobesson:
Do not stale. @momomaniac?

PR #987, opened 2020-01-30
Add fundraiser links

cbeauchesne:
Add fundraiser links based on hardcoded criterions.

### DIFF
--- a/src/js/get-fundraiser.js
+++ b/src/js/get-fundraiser.js
@@ -1,0 +1,88 @@
+const mapping = {
+  // 14274: { // France
+  //   url: ' https://www.helloasso.com/associations/rockclimber/formulaires/1/widget'
+  // },
+
+  14474: { // Scotland
+    url: 'https://www.7amax.co.uk/'
+  },
+
+  14151: { // Malta
+    url: 'https://maltaclimbingclub.org/bolt-fund/'
+  },
+
+  14125: { // Philippines
+    url: 'http://www.climbphilippines.com/boltfund.html'
+  }
+};
+
+const intersection = function(array1, array2) {
+  if (!array1 || !array2) {
+    return [];
+  }
+
+  return array1.filter(value => array2.includes(value));
+};
+
+export default function(document) {
+  // Only areas, waypoints, routes
+  if (!['r', 'w', 'a'].includes(document.type)) {
+    return null;
+  }
+
+  if (document.type === 'r') { // routes
+    // only mountain_climbing and rock_climbing activity
+    if (!intersection(document.activities, ['mountain_climbing', 'rock_climbing']).length) {
+      return null;
+    }
+
+    // only P1 and P1+ routes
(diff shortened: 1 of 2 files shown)


src/js/get-fundraiser.js
```diff
@@ -0,0 +1,62 @@
...
+
+export default function(document) {
+  // Only areas, waypoints, routes
+  if (!['r', 'w', 'a'].includes(document.type)) {
+    return null;
+  }
+
+  if (document.type === 'r') {
+    // for routes, only mountain_climbing and rock_climbing activity
+    if (intersection(document.activities, ['mountain_climbing', 'rock_climbing']).length === 0) {
```
brunobesson: `if (!intersection(document.activities, ['mountain_climbing', 'rock_climbing']).length)` is enough :)

src/js/get-fundraiser.js
```diff
@@ -0,0 +1,62 @@
...
+  // for waypoints, only climbing outdoor sites
+  if (document.type === 'w' && document.waypoint_type !== 'climbing_outdoor') {
+    return null;
+  }
+
+  const areas = [];
+  if (document.type === 'a') {
+    areas.push(document);
+  } else {
+    areas.push(...document.areas.filter((area) => area.area_type === 'range'));
```
brunobesson: `areas.push(...document.areas.filter((area) => ['range', 'admin_limits', 'country'].includes(area.area_type));`

src/js/get-fundraiser.js
```diff
@@ -0,0 +1,62 @@
...
+    areas.push(...document.areas.filter((area) => area.area_type === 'admin_limits'));
+    areas.push(...document.areas.filter((area) => area.area_type === 'country'));
+  }
+
+  for (const area of areas) {
```
brunobesson:
```js
return areas.map(area => mapping[area.document_id]).find((m) => m !== null)
```

could be enough?

src/js/get-fundraiser.js
cbeauchesne: Nope, the point of these three lines is to have first ranges, then admins_limit, then countries.

The logic behind is to display the item attached to the smallest entity. I also plan to include waypoints, and even routes in that logic : 

```
If an item is attached to the route => display it
else if an item is attached to it's main waypoint  => display it
else if an item is attached to an area of type range => display it
else if an item is attached to an area of type admin_limit => display it
else if an item is attached to an area of type country => display it
```

src/js/get-fundraiser.js
cbeauchesne: Yes, but IMO less readable.

src/js/get-fundraiser.js
cbeauchesne: 👍

src/js/get-fundraiser.js
brunobesson: up to you

PR #1044, opened 2020-03-29
Remove line number from .pot file

cbeauchesne:
Initially, I added line number in `.pot` file to help identify where a string come from.

Even if it help a little bit, it implies that any line addition/removal, or linting change lot of key, and makes `.pot` diffs hardly exploitable, as a lot of lines are considered as a change.

So, remove these line numbers.

### DIFF
--- a/tools/extract-messages.js
+++ b/tools/extract-messages.js
@@ -54,12 +54,13 @@ Result.prototype.toString = function() {
   return result;
 };
 
-function Process() {
+function Process(includeLineNumberInPositions) {
   // data is a simple key-value dict, where key
   this.data = {};
+  this.includeLineNumberInPositions = false;
 }
 
-Process.prototype.push = function(file, msgctxt, msgid) {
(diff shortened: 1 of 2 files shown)


tools/extract-messages.js
```diff
@@ -54,12 +54,13 @@ Result.prototype.toString = function() {
   return result;
 };
 
-function Process() {
+function Process(includeLineNumberInPositions) {
   // data is a simple key-value dict, where key
   this.data = {};
+  this.includeLineNumberInPositions = false;
```
brunobesson: Why don't you respect the function arg value?

cbeauchesne: Sorry, missed your comment :) 

Fat finger, i initially added the argument on the function for offering a CLI interface to enable it.

Then I changed my mind, low chance other user than me need this, and it's fast to change hard-coded `false` to `true` in the code. And I forgot to remove the argument in the prototype .... I will remove it in the next translations PR

Issue #1071, opened 2020-04-16
Feature: tag the lang of sections when the document language is not the same as the interface one.


lbesson:
https://forum.camptocamp.org/t/ag-2020-perspectives-devolution-du-topoguide/258857/23

Sections that are not in the same language as the interface should be tagged with the correct lang attribute.
Optionally, use translation tools to add a button for translating these sections

PR #1075, opened 2020-04-20
Add lang attribute for sections when document and interface langs differ

lbesson:
See #1071

I don't have any previous experience with vuejs, so maybe I'm not doing it the proper way.

I don't know if there are other places where the attribute should be added (maybe for document title?)

When this is ok, I can look for enabling translations (best UI for this to be defined)

### DIFF
--- a/src/views/document/utils/field-viewers/MarkdownSection.vue
+++ b/src/views/document/utils/field-viewers/MarkdownSection.vue
@@ -48,7 +48,7 @@
       lang() {
         const current_lang = this.$language.current;
         const document_lang = this.document ? this.document.cooked.lang : null;
-        return document_lang === current_lang ? undefined : document_lang;
+        return document_lang === current_lang ? undefined : this.$language.getIANALanguageSubtag(document_lang);
       }
     }
   };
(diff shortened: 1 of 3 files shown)


src/views/document/utils/field-viewers/MarkdownSection.vue
```diff
@@ -42,6 +42,14 @@
       return {
         visible: true
       };
+    },
+
+    computed: {
+      lang() {
+        const current_lang = this.$language.current;
+        const document_lang = this.document ? this.document.cooked.lang : null;
+        return document_lang === current_lang ? false : document_lang;
```
cbeauchesne, line 51:
```suggestion
        return document_lang === current_lang ? undefined : document_lang;
```

AFAIK, it's better to send `undefined` to HTML instead of `false` when we do not wan't to set it

brunobesson: If it's similar to angular, and I do think it is, then false will set the attribute's value, while undefined will remove it.

lbesson:
According to vuejs doc https://vuejs.org/v2/guide/syntax.html#Attributes I
would say that false undefined or null have the same effect


Le lun. 20 avr. 2020 à 21:25, Bruno Besson <notifications@github.com> a
écrit :

lbesson:
But I'm fine with undefined

cbeauchesne:
All good for me ! :rocket: @lbesson I let you merge it

PR #1090, opened 2020-04-25
Fix account saving 

### LINKED ISSUE #885: Bug: Submitting "My Account" form fails because of forum username
**Describe the bug**
In some cases submitting the ["My Account" form](https://www.camptocamp.org/account) (eg. when updating one's email address) fails because "Already used forum name".

**To Reproduce**
1. Go to https://www.camptocamp.org/account
2. Click on the "Save" button (no need to change anything in the form) => it works
3. Click again on the "Save" button => API returns a 400 error with message "Already used forum name"

**Expected behavior**
No error since the forum username is actually not changed.

**Analysis**
First time the form is submitted, field ``forum_username`` is not provided in the request parameters.
Second time it is submitted, field ``forum_username`` is provided (even though it was not changed and the form field not even focused on).
There is probably a bug in the UI, that should not push ``forum_username`` if unchanged.

Perhaps also a problem on the API side: it should be OK to update ``forum_username`` to its current value (even though it is not really relevant).

**Workaround**
The only way I have found to prevent the ``forum_username`` from being sent to the API is to log in / log out before submitting the form again (it is then "reset").

cbeauchesne:
Fix #885 (thanks a lot @asaunier for the analysis :pray:)

### DIFF
--- a/src/js/vue-plugins/user.js
+++ b/src/js/vue-plugins/user.js
@@ -97,8 +97,8 @@ export default function install(Vue) {
           is_profile_public,
           newpassword
         ).then(() => {
-          this.forumUsername = forum_username;
-          this.name = name;
+          this.forumUsername = forum_username === null ? this.forumUsername : forum_username;
+          this.name = name === null ? this.name : name;
           this.commitToLocaleStorage_();
         });
       },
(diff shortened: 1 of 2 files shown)


cbeauchesne:
@asaunier , can you check that it's fixed on https://c2corg.github.io/c2c_ui/fix-save-account/#/ ?

Issue #1140, opened 2020-05-10
Bug: distance in km displayed in meters [VTT]


desnoes:
**Describe the bug**
Field: "distance goudronnée"
The distance given in km in the form but displayed in meters with km as a unit.

**To Reproduce**
1. Go to https://www.camptocamp.org/routes/1158815/fr/doi-inthanon-depuis-chiang-mai

**Expected behavior**
The distance shall be displayed in km.

**Desktop (please complete the following information):**
 - Device: PC
 - OS: Windows
 - Browser : Mozilla

**Additional context**

PR #1141, opened 2020-05-10
Display MTB distance in km 

cbeauchesne:
issue #1140

### DIFF
--- a/src/components/generics/DocumentField.vue
+++ b/src/components/generics/DocumentField.vue
@@ -31,9 +31,9 @@
       {{ $gettext(value, field.i18nContext) }}
     </span>
 
-    <span v-else>{{ divisor ? Math.round(value / divisor) : value }}</span>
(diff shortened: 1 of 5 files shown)


src/components/generics/DocumentField.vue
```diff
@@ -31,9 +31,9 @@
       {{ $gettext(value, field.i18nContext) }}
     </span>
 
-    <span v-else>{{ divisor ? Math.round(value / divisor) : value }}</span>
-    <span v-if="showUnit && !field.skipSpaceBeforeUnit">&nbsp;</span>
-    <span v-if="showUnit">{{ unit || field.unit }}</span>
+    <span v-else>{{ divisor ? Math.round(value / divisor) : value }}</span><!--
```
brunobesson, line 34: there here to force spaces between spans?

cbeauchesne: To remove space between them, otherwise, the result is like `number-space-nbsp-space-unit`

PR #1147, opened 2020-05-11
Add support of TODO tags

asaunier:
Fixes https://github.com/c2corg/c2c_ui/issues/1172
API PR is https://github.com/c2corg/v6_api/pull/756

TODO:
* [x] add TODO button
* [x] add routes filter on TODO tag 
* [x] test document and user merging and check if tags are actually transfered
* [x] Fix codacy issue

Overview:

* the "todo" tag button in the document title ("star" icon as for following a contributor):

Screenshot from 2020-05-20 18-54-10

* "My Bookmarks" entry added to the user menu (name to be discussed I think):
Screenshot from 2020-05-20 18-54-46

### DIFF
--- a/src/js/apis/c2c/TagsService.js
+++ b/src/js/apis/c2c/TagsService.js
@@ -1,0 +1,30 @@
+function TagsService(api) {
+  this.api = api;
+}
+
+TagsService.prototype.add = function (document) {
+  if (document.type !== 'r') {
+    // As for now, only routes may be tagged.
+    return Promise.reject(new Error('Tags are not supported for this type of documents'));
+  }
+
+  const data = {
+    document_id: document.document_id,
+  };
+
+  return this.api.post('/tags/add', data);
+};
+
+TagsService.prototype.remove = function (document) {
+  const data = {
+    document_id: document.document_id,
+  };
+
+  return this.api.post('/tags/remove', data);
+};
+
+TagsService.prototype.get = function (document) {
+  return this.api.get('/tags/has/' + document.document_id);
+};
+
+export default TagsService;
--- a/src/js/apis/c2c/TagsService.js
+++ b/src/js/apis/c2c/TagsService.js
@@ -2,29 +2,20 @@ function TagsService(api) {
   this.api = api;
 }
 
-TagsService.prototype.add = function (document) {
-  if (document.type !== 'r') {
+TagsService.prototype.add = async function ({ type, document_id }) {
+  if (type !== 'r') {
     // As for now, only routes may be tagged.
-    return Promise.reject(new Error('Tags are not supported for this type of documents'));
+    throw new Error('Tags are not supported for this kind of document');
   }
-
-  const data = {
-    document_id: document.document_id,
-  };
-
-  return this.api.post('/tags/add', data);
+  return this.api.post('/tags/add', { document_id });
 };
 
-TagsService.prototype.remove = function (document) {
-  const data = {
-    document_id: document.document_id,
-  };
-
-  return this.api.post('/tags/remove', data);
+TagsService.prototype.remove = function ({ document_id }) {
+  return this.api.post('/tags/remove', { document_id });
 };
 
-TagsService.prototype.get = function (document) {
-  return this.api.get('/tags/has/' + document.document_id);
+TagsService.prototype.get = function ({ document_id }) {
+  return this.api.get('/tags/has/' + document_id);
 };
 
 export default TagsService;
--- a/src/js/apis/c2c/TagsService.js
+++ b/src/js/apis/c2c/TagsService.js
@@ -3,19 +3,25 @@ function TagsService(api) {
 }
 
 TagsService.prototype.add = async function ({ type, document_id }) {
+  checkDocumentType(type);
+  return this.api.post('/tags/add', { document_id });
+};
+
+TagsService.prototype.remove = function ({ type, document_id }) {
+  checkDocumentType(type);
+  return this.api.post('/tags/remove', { document_id });
+};
+
+TagsService.prototype.get = function ({ type, document_id }) {
+  checkDocumentType(type);
+  return this.api.get('/tags/has/' + document_id);
+};
+
+checkDocumentType = function (type) {
   if (type !== 'r') {
     // As for now, only routes may be tagged.
     throw new Error('Tags are not supported for this kind of document');
   }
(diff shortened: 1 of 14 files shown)


asaunier:
Oops! I have forgotten to create the new branch from https://github.com/c2corg/c2c_ui/tree/prettier instead of master...

src/js/apis/c2c/TagsService.js
```diff
@@ -0,0 +1,31 @@
+function TagsService(api) {
+  this.api = api;
+}
+
+TagsService.prototype.add = function(document) {
+  if (document.type !== 'r') {
+    // As for now, only routes may be tagged.
+    // TODO return an error message
+    return;
```
asaunier, line 9: Codacy doesn't like this because the returned value is not consistent with the other returned value.
Perhaps I could simply remove this test as the tags button is shown only for routes?

brunobesson:
> Oops! I have forgotten to create the new branch

Don't worry, prettier can be applied later on. All other commits have now been merged on master, so it is ok, maybe event better now to use master as reference branch.

src/js/apis/c2c/TagsService.js
brunobesson: what is returned in the route case?

(don't have time for a review tonight, but I may have a look tomorrow)

asaunier:
@brunobesson OK thanks, then I'll keep the current branch based on master.
Don't worry about reviewing now, I still have some work to do for this PR.

asaunier:
The PR is ready for review I think, even though I need to find a fix for the codacy complaint.

src/js/apis/c2c/TagsService.js
lbesson: you're returning a promise, so here you should return a `Promise.reject(new Error('not supported'))` or whichever you use for promises

lbesson:
I rebased the branch + linted + fixed the codacy issue

asaunier:
@lbesson Great! Thank you so much :)

src/js/apis/c2c/TagsService.js
```diff
@@ -0,0 +1,30 @@
+function TagsService(api) {
+  this.api = api;
+}
+
+TagsService.prototype.add = function (document) {
+  if (document.type !== 'r') {
+    // As for now, only routes may be tagged.
+    return Promise.reject(new Error('Tags are not supported for this type of documents'));
```
brunobesson, line 8:
```suggestion
    return Promise.reject(new Error('Tags are not supported for this type of document'));
```

src/js/apis/c2c/TagsService.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  const data = {
+    document_id: document.document_id,
+  };
+
+  return this.api.post('/tags/add', data);
+};
```
brunobesson, line 16: I would propose using object destructuring

```suggestion
TagsService.prototype.add = function ({type, document_id}) {
  if (type !== 'r') {
    // As for now, only routes may be tagged.
    return Promise.reject(new Error('Tags are not supported for this type of documents'));
  }
  return this.api.post('/tags/add', { document_id });
};
```

src/js/apis/c2c/TagsService.js
```diff
@@ -0,0 +1,30 @@
...
+
+TagsService.prototype.remove = function (document) {
+  const data = {
+    document_id: document.document_id,
+  };
```
brunobesson, line 21: Here also, I would use destructuring (althoug the argument type is then less clear because we use no jsdoc or typescript), or at least there is no need to instantiate `data` variable, just put this directly as argument for `post` method

src/views/SideMenu.vue
```diff
@@ -154,13 +154,13 @@ aside {
   bottom: 0;
```
brunobesson, line 154: not related to this PR, but obvious, so OK to include it

src/views/document/utils/TagsButton.vue
```diff
@@ -0,0 +1,64 @@
...
+        return this.$gettext('Unmark this route as to do');
+      } else {
+        return this.$gettext('Mark this route as to do');
+      }
+    },
+  },
+
+  created() {
+    if (this.canTag) {
+      c2c.tags.get(this.document).then((response) => {
```
brunobesson, line 36: you don't handle API errors. OK to fail quite silently, but maybe we should disable the star icon and add a error badge on it in this case, or something like that?

src/views/document/utils/TagsButton.vue
```diff
@@ -0,0 +1,64 @@
...
+        this.todo = response.data.todo;
+      });
+    }
+  },
+
+  methods: {
+    onClick() {
+      if (this.todo) {
```
brunobesson, line 44: here also, what should we do on error? which feedback?

src/views/document/utils/TagsButton.vue
```diff
@@ -0,0 +1,64 @@
...
+    tooltip() {
+      if (this.todo) {
+        return this.$gettext('Unmark this route as to do');
+      } else {
+        return this.$gettext('Mark this route as to do');
+      }
+    },
+  },
+
+  created() {
```
brunobesson, line 34: using async / await keywords would ease the reading IMHO, as chaingin is fairly simple

src/views/document/utils/TagsButton.vue
```diff
@@ -0,0 +1,64 @@
...
+    if (this.canTag) {
+      c2c.tags.get(this.document).then((response) => {
+        this.todo = response.data.todo;
+      });
+    }
+  },
+
+  methods: {
+    onClick() {
```
brunobesson, line 43: async / await could be used here too

src/views/document/utils/TagsButton.vue
```diff
@@ -0,0 +1,64 @@
...
+      }
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
+@import '@/assets/sass/variables.scss';
+
+.todo-button-yellow {
```
brunobesson, line 61: I don't remember how buttons are named all over the app, but I'd use descriptive words rather than a color for class name, e.g. `todo-button__active` (if we use something like BEM) or `.todo-button.active`

brunobesson (approved):

brunobesson:
I ended using toasts to reflect the error (thus adding a small new bulma plugin)

PR #1210, opened 2020-06-10
chore: sort imports with eslint

brunobesson:


src/App.vue
```diff
@@ -6,22 +6,24 @@
     <image-viewer ref="imageViewer" />
     <helper-window ref="helper" />
     <alert-window ref="alertWindow" />
+    <gdpr-modal />
```
cbeauchesne, line 9: Is it on purpose ?

brunobesson: no, I just corrected it. Forgot I had mixed it with some very preliminary stuff for GDPR. Sorry.

PR #1225, opened 2020-06-13
Rework image upload logic

### LINKED ISSUE #1224: Bug: "big" images are not resized
Currently, only JPEG images are preprocessed to handle rotation.
In addition to this, one should:

* handle "big" images and resize them
* ... not forgetting some special cases such as SVGs and panos

brunobesson:
* handle resizing in addition to orientation : no resizing for SVG, images under 2 MB. For other images, limit height to 2048px (thus allowing panos). The logic has flaws, but is quite simple and handles properly most cases.
* refactor code to use promises rather than callbacks, hopefully making it clearer
* handle errors

Fixes #1224

### DIFF
--- a/src/components/images-uploader/ImagesUploader.vue
+++ b/src/components/images-uploader/ImagesUploader.vue
@@ -253,6 +253,7 @@ export default {
 
     onUploadSuccess(document, image) {
       image.status = 'SUCCESS';
+
       Object.assign(image.document, document);
       this.computeReadyForSaving();
     },
@@ -260,7 +261,7 @@ export default {
     onUploadFailure(event, image) {
       image.percentCompleted = 100;
       image.status = 'FAILED';
-      image.errorMessage = event.message;
+      image.errorMessage = event?.message || this.$gettext('Image could not be processed');
     },
 
     computeReadyForSaving() {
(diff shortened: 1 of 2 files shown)


brunobesson:
Also fixes #1214

cbeauchesne:
For my knowledge, what was the issue on rotated images ? 

I'll do a review asap

cbeauchesne:
@brunobesson  If I understand well, the main difference is on these two lines : 

```js
  if (file.type !== 'image/svg+xml' && file.size > 2 * 1024 * 1024 && height > 2048) {
    options = { ...options, maxHeight: 2048 };
  }
```

The rest is some refactoring ?

brunobesson:
> The rest is some refactoring ?

Yes, mostly. But some of it is required for it to work properly. I can give you more details if needed.

brunobesson:
> For my knowledge, what was the issue on rotated images ?

I clearly don't only, maybe the blueimp-load-image version in prod is buggy. The code didn't change and is working now, not before... We could do some testing to understand why, not sure it is really usefull.

cbeauchesne:
> Yes, mostly. But some of it is required for it to work properly. I can give you more details if needed.

it's okay :) thank's a lot for the revamp, it's much clearer now

PR #1265, opened 2020-06-29
Clear association field on toggle

### LINKED ISSUE #1078: After selecting an autocomplete entry, the field should be cleared
From within the document edition, when trying to add an associated document. After selecting an entry fro mthe autcomplete suggestions, the text from the input should be cleared.

brunobesson:
Fixes #1078 

Couldn't manage to set focus back on field though. @cbeauchesne do you have any hint? `nextTick` or `setTimeout` did not do the trick.

### DIFF
--- a/src/components/generics/inputs/InputDocument.vue
+++ b/src/components/generics/inputs/InputDocument.vue
@@ -127,6 +127,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    clearInputOnToggle: {
+      type: Boolean,
+      default: false,
+    },
   },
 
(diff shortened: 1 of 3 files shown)


cbeauchesne:
I'm not sur to understand the point here. Is it about setting clearing the input when deselecting an item ?

brunobesson:
> I'm not sur to understand the point here.

What I would like is that when user select a document to associate (clicking on an item of the list), then the focus is set back to the input once association is created. This would be better from an UX point of view IMHO.

Issue #1131, opened 2020-05-07
Drop IE 11 support


brunobesson:
Once usage is sparse enough (let's say under 1% in analytics or caniuse for France), drop IE 11 support:

* [x] no more polyfills added by hand in `babel.config.json`
* [x] update browserslist in `package.json`
* [x] check that `toBlob` polyfill is still needed (currently also needed for Edge up to 18)
* [x] remove compatibility CSS, e.g. for `min()` and `max()` (see #1303)

Let stale bot warn us to check this again :)

brunobesson:
Do not stale. IE is now at a use level that enables us to remove it (<0.5%). Plus Microsoft started to take more actions to remove it (block some mainstream sites, etc).

PR #1304, opened 2020-07-29
Fix display of vertical thumbnail images into documents

### LINKED ISSUE #1303: Bug: Image thumbnails display
**Describe the bug**
Vertical images inserted into documents are often too big and blurry

**Screenshots**
Example 1 (https://www.camptocamp.org/routes/53823/fr/tete-des-fetoules-arete-w):
image

Example 2 (https://www.camptocamp.org/articles/269090/fr/les-coinceurs-protections-actives-coinceurs-mecaniques):
image

**Additional context**
Current css is:
```css
figure[c2c\:size='MI'] {
  width: 25vw;
  max-width: 400px;
  img {
    width: 25vw;
    max-width: 400px;
  }
}
```
I don't know the reasoning behind (maybe some kind of mix between not too blurry, but still be 'big enough' one the page), but IMHO, we should have the same max-width and max-height.
Current state is well for horizontal images, but bad for vertical ones

lbesson:


src/components/generics/Markdown.vue
```diff
@@ -222,12 +222,16 @@ figure[c2c\:role='embedded-figure'] {
 
 // medium image
 figure[c2c\:size='MI'] {
-  width: 25vw;
   max-width: 400px;
```
brunobesson, line 225: only IE11 in our current targeted browsers do not support `min()`. This compatibility layer should be removed when support is dropped. I added the mention in #1131

brunobesson:
Please rename your PR with a meaningfull title as this will be used for generating the changelog.

lbesson:
Isn't it the issue's title which is used?

brunobesson:
Indeed, I remove my comment as soon as I published it, but the mail was sent anyway.

PR #1312, opened 2020-07-30
Use debounce package

### LINKED ISSUE #141: Use debounce package
Regarding #138, which is probably a hack, include a strong and prefessional debounce method.

lbesson:
Closes #141


package.json
```diff
@@ -32,6 +32,7 @@
     "core-js": "3.6.5",
     "d3": "5.16.0",
     "file-saver": "2.0.2",
+    "lodash.debounce": "^4.0.8",
```
brunobesson, line 35: I'm not completely aware of the best solution, but I know at work we use lodash-es which allows for better tree shaking as the default lodash. Maybe the per module version is OK too (doesn't seem to be as up-to-date though).
Worth mentioning is we could also avoid depending on a lib and implements debounce easily with a few lines of code, e.g. https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_debounce

lbesson: I'm kinda agnostic on that subject. We could also use [debounce](https://www.npmjs.com/package/debounce) package, or one other.

In that case [lodash.debounce](https://www.npmjs.com/package/lodash.debounce) is small enough so that I don't think that it really is significant, but, well..

Even if it's unlikely to be updated anytime soon, maybe it's still better to rely on a 3rd party library (or use original code from @cbeauchesne, which is small)

brunobesson: OK, let's do this.

brunobesson: Please merge.

Issue #1332, opened 2020-08-08
Bug: XREPORT document - Double "Autres blessures" choice


anakreenskyrider:
**Describe the bug**
Dans le formulaire de rapport di'ncident il y a un double hoix "autres blessures" dans la partir "blessure antérieures"
voir screencshot
https://www.camptocamp.org/xreports/add/fr

Voir si ce choix est désiré !
![Sans titre](https://user-images.githubusercontent.com/49275256/89709107-387bd380-d97d-11ea-9d50-893cf90f5f26.png)

brunobesson:
@cbeauchesne why was it closed? fixed? not relevant? fixedwith translations? I could need to update extract tool too

cbeauchesne:
Fixed by #1375 no ?

brunobesson:
I had missed it, sorry.

PR #1375, opened 2020-09-14
fix(reports-fields): remove duplicated default values

joffreyverd:
Linked to the following issue: https://github.com/c2corg/c2c_ui/issues/1332

### DIFF
--- a/src/js/constants/common.json
+++ b/src/js/constants/common.json
@@ -218,7 +218,7 @@
     "orientation_types": ["N", "NE", "E", "SE", "S", "SW", "W", "NW"],
     "paragliding_ratings": ["1", "2", "3", "4"],
     "parking_fee_types": ["yes", "seasonal", "no"],
-    "previous_injuries": ["no", "previous_injuries_2"],
+    "previous_injuries": ["no", "previous_injuries_2", "previous_injuries_3"],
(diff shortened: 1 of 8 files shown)


cbeauchesne:
Hello @joffreyverd ,

Thank's a lot for your PR! 

Just, You do not need to touch translations files, they are generated automatically from our translation tool. Once we'll removed the property, it'll disappear automatically from those files

Side note, `common.json` is also generated automatically from the API DB model. Theorically, we should not touch it (really sorry, we definitely lack of docs 😞).

We can ship with this hack, but until it's not fixed on API side, we must be carefull to not overwrite this change.

cbeauchesne (changes requested):
Do not touch translations files

cbeauchesne:
oups, sorry, I was not clear enough:  `common.json` even if it's a hack, just keep it that way

joffreyverd:
Sorry for this misunderstanding!

cbeauchesne (approved):

cbeauchesne:
Thanks @joffreyverd !

cbeauchesne:
Released, thanks @joffreyverd  (btw, we have a developper group on the forum, useful for discussions with camptocamp association, if you're willing to, I can add you, simply give me your forum name).

joffreyverd:
@cbeauchesne here is my forum name: `laManteReligieuse`

PR #1471, opened 2020-11-09
Update photon.js

iamvdo:
The old URL (`.de` domain) no longer works

### DIFF
--- a/src/js/apis/photon.js
+++ b/src/js/apis/photon.js
@@ -4,7 +4,7 @@
 import BaseApi from '@/js/apis/BaseApi';
 
 function Photon() {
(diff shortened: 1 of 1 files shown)


iamvdo:
See https://github.com/komoot/photon/pull/508 for context

asaunier (approved):
Great, thanks!
Indeed the location search tool in prod no longer works, the requests to photo.komoot.de failing with a 301 Moved Permanently status (it seems the redirection does not work in our case).

cbeauchesne:
> Indeed the location search tool in prod no longer works,

I'll try to make a release today or tomorrow. Thanks @iamvdo !

cbeauchesne:
Released, thanks @iamvdo

PR #1478, opened 2020-11-12
Some files was missing on first prettier check

cbeauchesne:
I do not understand why those files was not cleaned when we introduced prettier Oo.

Fixing it now, otherwise it introduces lot of noise in other PRs, which is not very friendly with newcomers :)


brunobesson (approved):
I don't know either. husky and pretty-quick run linter only on staged files, but I thought I ran prettier on all files before. Maybe I forgot some?

PR #1479, opened 2020-11-13
Use nullish coalescing or default arg value when possible

brunobesson:
Just some code cleaning...


cbeauchesne:
Warning, `??` operator take the left part in case of empty string, or 0, where the `||`  take the right part. I know that I wasn't rigorous at all (sorry :cry: ). I'm pretty sure that I did not use it for `0` , but it's possible for an empty string. It worths a  check on each file :)

brunobesson:
I'm aware of that of course 👍 
I do think I checked every case and left out those that were dubious, but of course, I may have made some errors...

cbeauchesne (approved):

Issue #1449, opened 2020-10-25
Bug: Pagination buttons not obvious enough in list pages


asaunier:
**Describe the bug**
The current arrow buttons to go to the previous/next results page in the list pages are a bit too discrete and many users don't notice them, thinking the only results are the currently displayed ones.

**To Reproduce**
1. Go to https://www.camptocamp.org/routes?bbox=754884,5687198,774760,5717438

<img width="514" alt="Capture d’écran 2020-10-25 à 20 47 09" src="https://user-images.githubusercontent.com/1192331/97117487-5fd47800-1704-11eb-9003-14a6cdbc3d37.png">

**Expected behavior**
Making it more obvious there are other results pages.

cbeauchesne:
I totally agree it's not easy to understand for newcomers.

Coding the solution is not hard. But finding it is not easy. Things to consider : 

* must be readable on both desktop and mobile (we can have two different solutions)
* Knowing the total count of document is important
* Vertical space is scare

And if the solution is more than a simple change in design (meaning it asks a consequent dev effort, and present a breaking change in user experience), a short presentation on forum is needed.

ataraxido:
Hello,

What do you think to modify the behavior of this feature and to go this way :

- "1-30 sur 45701" < > becomes "45701 results" or "45701 routes"
- add an infinite scroll ( ie when the user scrolls up to the bottom of the list it automatically append 30 following results )

It could also be done in an no-auto way with an overlay "load more" button at the bottom of the list

cbeauchesne:
The infinite scroll comes with it pitfalls. I've in mind this: 

* it will make the map not very user friendly, as icon will appears when you scroll down
* with map on mode "view only result on this map", any move of the map will reset the list to the top
* it makes hard to build an URL with an offset (yes I know, it's an edge-case usage)

I'm not saying it's not the good solution, but it's not a silver bullet. If we choose it, it'll requires a lot of attention.

> It could also be done in an no-auto way with an overlay "load more" button at the bottom of the listIt could also be done in an no-auto way with an overlay "load more" button at the bottom of the list

It'll eat some vertical space, scare ressource.

By the way, the issue is about making the button more explicit. Simply transform them in a more button-design like could do the job ?

ataraxido:
I got your point, I didn't think of the map since I'm not using it much but you're right.

ataraxido:
First remark : it's way more an enhancement rather than a bug

Also here my (working) proposition :

[...]

Issue #499, opened 2019-04-09
Find a way to display figures about results


cbeauchesne:
Actually, available data about documents counts are :  

* the number of page N (capped to 99, which means "lot of results")
* and, implicitly, the total count of result : N*30

It may be usefull to be more explicit with the fact that results on map are partial. But I've actually no idea how to do this, without spamming the UI. 

Also, power users may needs to have explicitly the number of results (N*30, not capped). Same constat : it's a specific use case, and it would be preferable to find a way to show this, without anyoing generals user who don't care to know that camptocamp contains 543,567 images or 234,432 waypoints.

In previous UI, the solution was to show a label like "90-120/32,143"

IMO, it's way to much visible. Suggestions are welcome.

cbeauchesne:
Community decision : https://forum.camptocamp.org/t/adoption-de-la-nouvelle-interface-graphique-du-site-camptocamp-org/239741/147?u=charlesb

PR #1483, opened 2020-11-16
Pagination

ataraxido:
For https://github.com/c2corg/c2c_ui/issues/1449


cbeauchesne:
Demo app: https://c2corg.github.io/c2c_ui/pagination/#/outings?qa=draft,great&bbox=-2127335,3118903,3626760,8439377

I'll have a look tonight

asaunier:
Capture d’écran 2020-11-16 à 19 09 49

Nice :)
Cool to be able to choose the offset!

A few remarks though:
* Will users understand at first sight what the "30/60/90" dropdown is for?
* Will users understand at first sight what "3 sur 20+" stands for? At first I thought it was the number of outings, but it seems to be the number of pages, right?
* I don't think normal users know what "20+" means :) 
* Long dropdowns are not very convenient: you have to open them and scroll it with your cursor every time you want to go to the next page (links such as "next" are much faster to use)
* It's impossible to have more than 20 pages of results, right? Actually you can go to page 21 by clicking on ">" but it's not consistent with the dropdown, thus confusing.

My suggestions (I have not though much about it):
* Explicitly indicate the number of results: eg. "123456 outings" (as you did => great!) => this way the user see there is more results than what's actually displayed
* Move the pagination at the bottom of the page => you browse the page from top to bottom, if no result matches what you want when arrived at the bottom, you click to get the next page. It's a pain in the neck to have to scroll back up to go to the next page. Why not having the pagination at the top of the page but then it would cool to duplicate it at the bottom.
* Long dropdowns are not convenient (see my comment above), I would suggest to rather use links/buttons.
* To make the pagination obvious, what about using the design of well-known pages (eg. google):
Capture d’écran 2020-11-16 à 19 23 00

To keep things simple and space-saving what about using a block like this:
```
<  91-120 sur 230008 sorties >
```
OK it's very close from the existing:
Capture d’écran 2020-11-16 à 19 25 55
but perhaps it's enough?

ataraxido:
Honestly I'm _a bit_ surprised by those UI/UX remarks : the issue has been opened by you 20 days ago ... Why not suggesting/discussing this before someone works on it ?

> Will users understand at first sight what the "30/60/90" dropdown is for?
Maybe not indeed, I just tried to save horizontal space, which wording/display would you suggest ?

> Will users understand at first sight what "3 sur 20+" stands for?
Maybe not at first sight, but at first use I'm quite sure of it. It's a very common use-case.

> I don't think normal users know what "20+" means 
I was not sure neither but I had no better idea. I just wanted to show that the result is capped. Any suggestion ?

> Long dropdowns are not very convenient
I agree, that's why I capped the result to 20 pages

> Links such as "next" are much faster to use because they are always at the same place => no need to move the cursor between clicks
I strongly disagree, if you want to switch to page #18 you'll have to click 'Next' 17 times whereas with this list it's 2 clicks.

> It's impossible to have more than 20 pages of results, right? Actually you can go to page 21 by clicking on ">" but it's not consistent with the dropdown, thus confusing.
When going at page #22 for example using 'Next' then nothing is selected on the 20+ dropdown, so for me it's consistent even if I agree a QA manager could be sad of it :)

About your suggestions :+1: 

- moving the pagination to the bottom of the page : I would totally agree but according to @cbeauchesne the vertical space is something to take much care of.

> To make the pagination obvious, what about using the design of well-known pages (eg. google)
This doesn't allow to change the number of results per row which according to me the main issue for such outings/routes listings

asaunier:
> Honestly I'm a bit surprised by those UI/UX remarks : the issue has been opened by you 20 days ago ... Why not suggesting/discussing this before someone works on it ?

About a month ago, a friend of mine sent me an email because she couldn't find what she wanted in the search pages. I suspected she had not noticed the pagination stuff and opened an issue to keep track of this UI/UX problem.
But spent no more time thinking of it. It's better to spend it when someone actually tackles the issue as you do.

Issues at c2corg are often draft issues with no detailed specs. 

> I just tried to save horizontal space, which wording/display would you suggest ?

To be explicit enough that could be "Show 30/60/90 results".
If it takes too much space on small devices, perhaps we should consider not showing this option on such devices as it is not that essential.

> Will users understand at first sight what "3 sur 20+" stands for?

In addition it's quite a lot of repeated text (" sur 20+"). What about only having the page number in the dropdown with the surrounding text: "Page ()/20"?
If more than 20 pages, instead of "/20+" we could also use humour like this "Page ()/many" :)

> I strongly disagree, if you want to switch to page #18 you'll have to click 'Next' 17 times whereas with this list it's 2 clicks.

You're right!
But would people really want to see page 18? That's the same than on Google, I guess people only look at the 2 or 3 first pages, don't you think?

> I would totally agree but according to @cbeauchesne the vertical space is something to take much care of. But yeah, it would make much sense.

With the current design, the filters form and pagination are fixed at the top of the page.
But does this really make sense? When you browse the results, it's better to see the results than the form.
What about "unfixing" the filters form and the pagination and have them disappear when scrolling down?
This way no problem having the pagination appear at the bottom when reaching the end of the list.
By the way, the pagination and number of results are parts of the results (on the contrary of the filters form) => that would make sense to move them inside the results area (even if it is decided that the filter form should remain fixed).

> This doesn't allow to change the number of results per row which according to me the main issue for such outings/routes listings

OK, I just wanted to say that users may be more used to raw/simple design as the google approach.
It's not contradictory with the offset tool: the action/URL behind the arrows/next/previous links or the figures links would simply change depending on the selected offset, as it's the case with the dropdown.

Again, I'm not against the dropdown approach, just challenging it because the "flat" approach might be good too.

cbeauchesne:
> To make the pagination obvious, what about using the design of well-known pages (eg. google):

For the reccord, this solution was the very first one, but it was a no-go from the users. Let me find the discussion

cbeauchesne:
https://github.com/c2corg/c2c_ui/issues/499

I've strongly pushed for the google-like pagination. It was a fail :)

ataraxido:
> To be explicit enough that could be "Show 30/60/90 results".
If it takes too much space on small devices, perhaps we should consider not showing this option on such devices as it is not that essential.

Okay; Maybe only "Show" since we know what we are showing.

> If more than 20 pages, instead of "/20+" we could also use humour like this "Page ()/many" :)
I though of "many" instead of 20+, I just tried to avoid new translation but yeah probably better.

> In addition it's quite a lot of repeated text (" sur 20+"). What about only having the page number in the dropdown with the surrounding text: "Page ()/20"?
It's doesn't change things much but I don't like the display, but if you prefer this way fine.

> But would people really want to see page 18? That's the same than on Google, I guess people only look at the 2 or 3 first pages, don't you think?
No, actually here is a very very different thing than on Google. Googling something isn't very accurate, the diversity of results is very wide whereas here when you are searching for routes you find ... routes :) And the possibilities to filter on c2c are way more developed than on Google, so I think it's a very bad idea to compare with an online search engine.

Be sure people aren't looking at 2-3 top pages only : I found myself browsing many numerous results a lot of time. Maybe more on routes than on outings. But typically I am looking for my next mixed outing so I search for mixed routes with grade F-PD+ and I get tons of results that I browse :)

> It's not contradictory with the offset tool: the action/URL behind the arrows/next/previous links or the figures links would simply change depending on the selected offset, as it's the case with the dropdown.

Actually the dropdown doesn't change the offset of the query but it's limit, I think you misunderstood the way it work ( which isn't a good POC ).

I agree with putting the pagination at the bottom within/close to the results area. I had this in mind since the beginning. @cbeauchesne should we than crunch a bit of vertical space in order to do that ?

asaunier:
> if you prefer this way fine.

@ataraxido Sorry I didn't want to appear directive. Just debating again. No matter what I prefer :) Perhaps other people think different.

> I've strongly pushed for the google-like pagination. It was a fail :)

@cbeauchesne Only people debating as we do now. At the end we still have a page where "not-power" users don't figure out there are other pages of results :P

cbeauchesne:
I've some opinions, but it'll be only a one-man opinion (and this project shows me more than once that my opinion may diverges a lot from other users' opinion).

If we change the ergonomy here, I think it would be a good idea to gather some feedback from users. Otherwise, we may face some negative reactions, which is the worse fort both users, and developer's motivation. 

@ataraxido , If you're ok, I can handle this discussion (I'm pretty used to animate them).

cbeauchesne:
> Honestly I'm a bit surprised by those UI/UX remarks : the issue has been opened by you 20 days ago ... Why not suggesting/discussing this before someone works on it ?

Actually, we do not have a clear workflow about issue creation/resolution. When it's simply a bug fix (and yes, this issue was not a bug, more a design enhancement) or a minor design change, we usually let the dev choose the best solution, and the PR review handles everything).

But when it's about new features, major design change, or ergonomy change, most of the time nothing is specified in issues, and we discuss it on the forum (private to association for some specific subject, public one for most of the subject). We even have a procedure to debate : https://www.camptocamp.org/articles/1008305/fr/-gestion-c2c-micro-procedure-de-prise-de-decision when we are not able to find a consensus.

ataraxido:
@cbeauchesne for sure you can if you want ; I just hope it will bring us more than "I like" or "I dislike" without any argument or detailed explanation.

I have a web-software professional background ( like you it seems ) but no real experience of open-source project, I read the discussions you attached and has been quite surprised of how it leads to decision but I guess I have to get to know more this open-source associative projects.

Expressing it as a climber point of view : of course sometimes you dislike a route or think it's been poorly designed but since bolting guys are usually way more experienced than climbers and also spent their time freely bolting the route then it (quite) never leads the route to be removed/modified.

cbeauchesne:
Discussion : https://forum.camptocamp.org/t/boutons-de-pagination/270555

For the record, I found a screenshot of the pagination I did initially : https://forum.camptocamp.org/uploads/default/original/2X/2/26f858edf49700cdb6ecb7cc34ea5dddf0e34267.jpg

asaunier:
@ataraxido Thank you very much for the changes you did! What I see at https://c2corg.github.io/c2c_ui/pagination/#/outings?qa=draft,great&bbox=740801,5809986,758305,5831923&offset=630&limit=60&act=skitouring looks like a great improvement to me!

AntoineMarnat:
Hello,

I did not read the whole discussion (yet), and just looked at the demo: it is nice !

My suggestion would be to set the default value of shown items to 60.

cbeauchesne:
> My suggestion would be to set the default value of shown items to 60.

As long as we do not have any tool to monitor the API health (memory and CPU), please keep it to 30, don't want to break the API 🤔

ataraxido:
What is the usual time span of such open discussions ?

cbeauchesne:
I've made a small recap of discussion: https://forum.camptocamp.org/t/boutons-de-pagination/270555/17?u=charlesb

As it seems to be a good consensus, the next is to propose a solution, and see reactions :

* if they are globally positive => release
* otherwise, gather more feedback

Side technical note about the display with list (table) : to make the table vertically scrollable, you have to know the available vertical size. It's a PITA, and it has not be done very well (newbie inside). Example of terrible code here: 

https://github.com/c2corg/c2c_ui/blob/master/src/views/documents/DocumentsView.vue#L264

If you choose the solution to move the pagination tool to the bottom, please be careful with all possible displays. And if you're a star with flexbox, there is a lot of room for improvement here ;)

ataraxido:
Okay, I did everything that you gathered on the forum synthesis except to move the pagination at the bottom.

I finally noticed that the results "frame" has a fixed height and so the upper filters/buttons are always visible : so it's not really mandatory to have the pagination at the bottom. I think it would still be consistent to have them as the bottom but indeed ( I did some tests ) it crunchs some vertical space and when using the map is (according to me) let's convenient than to have it on top.

Let see what users think of this.

cbeauchesne:
> ...it crunchs some vertical space...

I've rage quit on this part 😄 

I'll post a message on the forum tonight.

ataraxido:
Would you like me to do it Charles ? I've no problem and plenty of time.

cbeauchesne:
> Would you like me to do it Charles ? I've no problem and plenty of time.

Sure, TYVM!

cbeauchesne (approved):

PR #1489, opened 2020-11-17
Optimize some images

brunobesson:
* we have huge background images for home, serac and topoguide views. Let's try to optimize those images size + use smaller one when possible
* mainly for serac, we have a lot of static images, which can benefit from being served with different formats, add webp and avif


src/views/static-views/SeracView.vue
```diff
@@ -125,10 +125,18 @@ export default {
 
 <style scoped lang="scss">
 .serac-view {
-  background-image: url('~@/assets/img/backgrounds/cermygian-crevasse.jpg');
+  background-image: url('~@/assets/img/backgrounds/cermygian-crevasse-800w.jpg');
   background-size: cover;
   background-position-y: 50%;
+  background-position-x: center;
```
ataraxido, line 131: just background-position: center; ?

ataraxido (commented):
Small improvement suggestion

src/views/static-views/SeracView.vue
brunobesson: right 👍

PR #1504, opened 2020-11-24
Add some JSON-LD

brunobesson:
See https://forum.camptocamp.org/t/meta-donnees/258983
Let's try to introduce some info.

@cbeauchesne don't know if there is a way to reference the server properly rather than putting absolute paths?


cbeauchesne:
Noob question: how to test it ? 

:D

brunobesson:
https://search.google.com/test/rich-results

brunobesson:
(I made one more fix to be propagated to github pages)

cbeauchesne (approved):

PR #1508, opened 2020-11-25
Add open graph meta tags

brunobesson:


src/js/vue-plugins/gettext-plugin.js
```diff
@@ -177,6 +177,31 @@ export default function install(Vue) {
...
+            return 'it_IT';
+          case 'de':
+            return 'de_DE';
+          case 'es':
+            return 'es_ES';
+          case 'zh_CN':
+            return 'zh_CN';
+          default:
+            // eslint-disable-next-line no-console
+            console.error(`Unexpected langauage: ${lang}`);
```
AntoineMarnat, line 200: typo ;)

brunobesson: 🙂

src/js/vue-plugins/gettext-plugin.js
```diff
@@ -172,7 +172,32 @@ export default function install(Vue) {
...
+      getIsoLanguageTerritory(lang) {
+        switch (lang) {
+          case 'fr':
+            return 'fr_FR';
+          case 'en':
+            return 'en_UK';
+          case 'ca':
+            return 'ca_ES';
+          case 'eu':
+            return 'eu_E';
```
ataraxido, line 189: typo ? does not exist

src/views/document/utils/document-view-mixin.js
```diff
@@ -67,6 +68,12 @@ export default {
         },
       ];
     },
+    meta: function () {
+      if (!this.document) {
+        return null;
+      }
+      return this.documentOpenGraph();
```
ataraxido, line 75:
```suggestion
      if (this.document) {
        return this.documentOpenGraph();
      }
      return null;
```

src/js/vue-plugins/gettext-plugin.js
brunobesson: 👍

src/views/document/utils/document-view-mixin.js
brunobesson: I'd rather keep the logic the other way: first return fast with excluding conditions, then expand the main logic (although it's simple here)

Issue #1505, opened 2020-11-24
Feature: encourage filling history of a route for apropriate activity


AntoineMarnat:
Camptocamp is aiming to have the best information about the history of routes - and very good at it. I suggest the following to be even better

**Describe the solution you'd like**
For the following activities: snow_ice_mixed, mountain_climbing, rock_climbing, ice_climbing, via_ferrata, slacklining
If the history field is empty, instead of hiding it, show it with the text " You know the history of the route ? Fill it in ! " as a link to the edition page, open at the history section.

It follows the behavior for cotation, see:
https://www.camptocamp.org/routes/1261111/

If technically no too tricky, one could also add skitouring under the condition that the cotation is higher than 5.1:  trat=5.1,5.6

asaunier:
> If the history field is empty, instead of hiding it, show it with the text " You know the history of the route ? Fill it in ! " as a link to the edition page, open at the history section.

Should not be displayed when printing

AntoineMarnat:
Hello,

I would be happy to try to give it a go, but I would need some help. In particular, I am not familiar at all with the tree of the different files.

As far as I understand, the changes must be done in this file:
https://github.com/c2corg/c2c_ui/blob/master/src/views/document/RouteView.vue
Example of the rating : lines 27-29:

> <label-value :label="$gettext('ratings')">
>                 <document-rating v-if="$documentUtils.hasRating(document)" :document="document" show-helper />
>                 <edit-link v-else :document="document" :lang="$user.lang" />

and changing line 80

>          <markdown-section :document="document" :field="fields.route_history" />

Unfortunatly, I can't find where "hasRating" is defined ... If someone can hint me where the printing setup is ?

brunobesson:
Hi. The `hasRating` function is defined in `src\js\vue-plugins\document-utils.js`. It is developped as a Vue plugin which adds a global property `$documentUtils`, which you can thus access with `this` keyword.

What you need to do then:
* either display the `MarkdownSection` component if route history is filled or deosn't match your criteria, otherwise display another component which will present your text
* or customize the `MarkdownSection` component to handle specific cases.

Which one is best depends on the complexity and structure of the modifications, but I assume it will be best to modify the component.

cbeauchesne:
It may be something like transforming 

https://github.com/c2corg/c2c_ui/blob/master/src/views/document/RouteView.vue#L80

```html 
<markdown-section :document="document" :field="fields.route_history" />
```

into 

```html 
[...]

PR #1510, opened 2020-11-27
Update RouteView.vue:  History field  #1505

AntoineMarnat:
My first try, 
not sure how v-translate works to push to transifex and modify the file in '/src/translations'.
not sure how '`<edit-link>`' is defined

### DIFF
--- a/src/views/document/RouteView.vue
+++ b/src/views/document/RouteView.vue
@@ -78,12 +78,13 @@
         <div class="box">
           <markdown-section :document="document" :field="fields.summary" />
           <markdown-section v-if="locale.route_history" :document="document" :field="fields.route_history" />
-            <div v-else-if="activities.includes('snow_ice_mixed') || activities.includes('mountain_climbing') || activities.includes('rock_climbing') || activities.includes('ice_climbing') || activities.includes('via_ferrata') || activities.includes('slacklining') || activities.includes('skitouring') && ['5.1', '5.2', '5.3', '5.4', '5.5','5.6'].includes(doc.ski_rating))" class="missing-history-banner no-print"
-              <edit-link v-if="isEditable" :document="document" :lang="lang">
-               <v-translate/> Missing history/>
-              </edit-link>
-            </div>
-            <div v-else disable </div>
+             <div v-else-if="PromoteHistory" class="missing-history-banner no-print" />
+               <edit-link v-if="isEditable" :document="document" :lang="lang" v-translate/> 
+                  "missing_history"
+                </edit-link>
+             <div v-else>
+               disable
+             </div>
           <markdown-section :document="document" :field="fields.description" />
           <markdown-section :document="document" :field="fields.slackline_anchor1" />
           <markdown-section :document="document" :field="fields.slackline_anchor2" />
@@ -125,6 +126,7 @@
 
 <script>
 import documentViewMixin from './utils/document-view-mixin';
+import isEditableMixin from './is-editable-mixin';
 
 export default {
   mixins: [documentViewMixin],
@@ -186,6 +188,25 @@ export default {
 
       return result;
     },
+    documentType() { // is-editable mixin needs this property
+        return this.$documentUtils.getDocumentType(this.document.type);
+      },
+    PromoteHistory() {
+        const result = {};
+        const doc = this.document;
+        const activities = doc.activities ?? [];
+        const history_worth_activities = ['snow_ice_mixed', 'mountain_climbing', 'rock_climbing', 'ice_climbing', 'via_ferrata', 'slacklining'];
+        
+        for act in history_worth_activities{
+        if (activities.includes(act)){
+          result = true
+          }
+        }
+        if (activities.includes(skitouring) && ['5.1','5.2','5.3','5.4','5.5'].includes(doc.ski_rating)){
+          result = true
+          }
+        return result;
+    }
   },
 };
 </script>
--- a/src/views/document/RouteView.vue
+++ b/src/views/document/RouteView.vue
@@ -77,7 +77,13 @@
 
         <div class="box">
           <markdown-section :document="document" :field="fields.summary" />
-          <markdown-section :document="document" :field="fields.route_history" />
+          <markdown-section v-if="locale.route_history" :document="document" :field="fields.route_history" />
+            <div v-else-if="activities.includes('snow_ice_mixed') || activities.includes('mountain_climbing') || activities.includes('rock_climbing') || activities.includes('ice_climbing') || activities.includes('via_ferrata') || activities.includes('slacklining') || activities.includes('skitouring') && ['5.1', '5.2', '5.3', '5.4', '5.5','5.6'].includes(doc.ski_rating))" class="missing-history-banner no-print"
+              <edit-link v-if="isEditable" :document="document" :lang="lang">
+               <v-translate/> Missing history/>
+              </edit-link>
+            </div>
+            <div v-else disable </div>
           <markdown-section :document="document" :field="fields.description" />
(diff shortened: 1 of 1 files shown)


src/views/document/RouteView.vue
```diff
@@ -77,7 +77,13 @@
 
         <div class="box">
           <markdown-section :document="document" :field="fields.summary" />
-          <markdown-section :document="document" :field="fields.route_history" />
+          <markdown-section v-if="locale.route_history" :document="document" :field="fields.route_history" />
+            <div v-else-if="activities.includes('snow_ice_mixed') || activities.includes('mountain_climbing') || activities.includes('rock_climbing') || activities.includes('ice_climbing') || activities.includes('via_ferrata') || activities.includes('slacklining') || activities.includes('skitouring') && ['5.1', '5.2', '5.3', '5.4', '5.5','5.6'].includes(doc.ski_rating))" class="missing-history-banner no-print"
```
brunobesson, line 81: the logic should be extracted into a function. it's far too complicated to stays in the template.

src/views/document/RouteView.vue
```diff
@@ -77,7 +77,13 @@
...
+              <edit-link v-if="isEditable" :document="document" :lang="lang">
+               <v-translate/> Missing history/>
```
brunobesson, line 83: There is a typo here, your tag is not properly opened and closed

src/views/document/RouteView.vue
```diff
@@ -77,7 +77,13 @@
...
+              </edit-link>
+            </div>
+            <div v-else disable </div>
```
brunobesson, line 86: syntax error too

src/views/document/RouteView.vue
```diff
@@ -77,7 +77,13 @@
 
         <div class="box">
           <markdown-section :document="document" :field="fields.summary" />
-          <markdown-section :document="document" :field="fields.route_history" />
+          <markdown-section v-if="locale.route_history" :document="document" :field="fields.route_history" />
+            <div v-else-if="activities.includes('snow_ice_mixed') || activities.includes('mountain_climbing') || activities.includes('rock_climbing') || activities.includes('ice_climbing') || activities.includes('via_ferrata') || activities.includes('slacklining') || activities.includes('skitouring') && ['5.1', '5.2', '5.3', '5.4', '5.5','5.6'].includes(doc.ski_rating))" class="missing-history-banner no-print"
+              <edit-link v-if="isEditable" :document="document" :lang="lang">
```
brunobesson, line 82: you need to use the is-editable mixin to use `isEditable` property

src/views/document/RouteView.vue
brunobesson, line 81: If there is no styling applied, I'm pretty sure it will look ugly :)

cbeauchesne: @AntoineMarnat , just run `npm run lint` before committing, it'll  clean your code.

brunobesson: What kind of IDE are you using @AntoineMarnat? If you use VS Code (which I would recommend), you can install the prettier plugin and configure it to format on save. This would ease your burden.

src/views/document/RouteView.vue
AntoineMarnat: OK, done !

src/views/document/RouteView.vue
AntoineMarnat: Hopefully done ?

AntoineMarnat: I tried to use Github desktop and now VS Code, but somehow I fail to push commits ... so I edit directly on github website :/

src/views/document/RouteView.vue
```diff
@@ -78,12 +78,13 @@
...
           <markdown-section v-if="locale.route_history" :document="document" :field="fields.route_history" />
-            <div v-else-if="activities.includes('snow_ice_mixed') || activities.includes('mountain_climbing') || activities.includes('rock_climbing') || activities.includes('ice_climbing') || activities.includes('via_ferrata') || activities.includes('slacklining') || activities.includes('skitouring') && ['5.1', '5.2', '5.3', '5.4', '5.5','5.6'].includes(doc.ski_rating))" class="missing-history-banner no-print"
-              <edit-link v-if="isEditable" :document="document" :lang="lang">
-               <v-translate/> Missing history/>
-              </edit-link>
-            </div>
-            <div v-else disable </div>
+             <div v-else-if="PromoteHistory" class="missing-history-banner no-print" />
+               <edit-link v-if="isEditable" :document="document" :lang="lang" v-translate/> 
+                  "missing_history"
```
brunobesson, line 83: Do not use quotes.
Also, the i18n for c2c_ui is using mostly english sentences to esae the translation work. You could .e.g use `History is missing. Edit the route to complete the information.`

src/views/document/RouteView.vue
```diff
@@ -78,12 +78,13 @@
...
+                </edit-link>
+             <div v-else>
+               disable
```
brunobesson, line 86: what should this produce? it's gonna write `disable` on screen?

src/views/document/RouteView.vue
```diff
@@ -186,6 +188,25 @@ export default {
 
       return result;
     },
+    documentType() { // is-editable mixin needs this property
+        return this.$documentUtils.getDocumentType(this.document.type);
+      },
```
brunobesson, line 193: there are linting problems, but we can fix that later

src/views/document/RouteView.vue
```diff
@@ -186,6 +188,25 @@ export default {
...
+    PromoteHistory() {
```
brunobesson, line 194: do not use upper case for function names.
It's a test function, so better to be named e.g. `shouldPromoteHistory`

src/views/document/RouteView.vue
```diff
@@ -186,6 +188,25 @@ export default {
...
+        const result = {};
```
brunobesson, line 195: you declare result as an object but ater on it's a boolean. anyway, you don't need to declare it, just return. E.g.

```js
if (activities.includes(act)) {
  return true;
}
```

src/views/document/RouteView.vue
```diff
@@ -186,6 +188,25 @@ export default {
...
+        const doc = this.document;
+        const activities = doc.activities ?? [];
+        const history_worth_activities = ['snow_ice_mixed', 'mountain_climbing', 'rock_climbing', 'ice_climbing', 'via_ferrata', 'slacklining'];
```
brunobesson, line 198: those consts could be declared outside of the vue export so it's not computed each time the function is called (I know it's a copy & paste and could be improved in other places too)

src/views/document/RouteView.vue
```diff
@@ -186,6 +188,25 @@ export default {
...
+        
+        for act in history_worth_activities{
+        if (activities.includes(act)){
+          result = true
+          }
+        }
+        if (activities.includes(skitouring) && ['5.1','5.2','5.3','5.4','5.5'].includes(doc.ski_rating)){
```
brunobesson, line 205: `skitouring` is not a variable but a string.

brunobesson (changes requested):
Thanks for all the work.

You should really not try to do PR without testing it in a local environment though 🙂 

There are IMHO many little things to fix, but this is quite easy.

I have a bigger problem with 1/ the UI part (e.g. the disabled value) and 2/ the logic applied.

As for the logic, it should:
* display the history block if present
* if not and criteria are met o display the edit hint, they do it
* otherwise nothing is displayed (no v-else)

cbeauchesne:
A new PR from @AntoineMarnat will come soon

PR #1513, opened 2020-11-30
Add edit links when history is missing for appropriate activities #1505

AntoineMarnat:


src/components/generics/links/EditLink.vue
```diff
@@ -21,6 +21,10 @@ export default {
       type: String,
       required: true,
     },
+    showEvenIfNotLogged: {
+      type: Boolean,
+      default: false,
+    },
```
brunobesson, line 27: The semantic is correct, but it feels a bit long. What about e.g. `showAlways`?

src/views/document/RouteView.vue
```diff
@@ -192,11 +218,10 @@ export default {
...
 }
-
 @media print {
   .protection-area-info {
     margin: 0rem !important;
     padding: 0rem !important;
   }
 }
-</style>
+</style>
```
brunobesson, line 227: please leave empty line at end of file

src/views/document/RouteView.vue
```diff
@@ -180,6 +185,27 @@ export default {
 
       return result;
     },
+    showMissingHistoryBanner() {
+      const doc = this.document;
+      const activities = doc.activities ?? [];
+      const history_worth_activities = [
```
brunobesson, line 191: Please use camelCase notation for variables. We have some properties with underscores, but that's because they come directly from the api.

brunobesson, line 191: to avoid computing the array each time, you could move the declaration out of the default export (agreed, very small perf improvement)

src/views/document/RouteView.vue
```diff
@@ -77,7 +77,12 @@
 
         <div class="box">
           <markdown-section :document="document" :field="fields.summary" />
-          <markdown-section :document="document" :field="fields.route_history" />
+          <markdown-section v-if="locale.route_history" :document="document" :field="fields.route_history" />
+          <div v-else-if="showMissingHistoryBanner" class="notification is-info no-print">
+            <edit-link :document="document" :lang="lang" show-even-if-not-logged>
+              History is missing, please provide it if you have information.
+            </edit-link>
+          </div>
```
brunobesson, line 85: Couldt it be more consistent if the notifcation box was displayed below a history section? Currently, there is no history section.

brunobesson (approved):
I like it a lot more 👍

src/views/document/RouteView.vue
AntoineMarnat: OK, is it always mandatory to do so ?

src/views/document/RouteView.vue
AntoineMarnat: I do not know. I would say that it is only an advertising for editing, not an information for the random reader: maybe with a History section it uses too much space ?

AntoineMarnat:
Question: how and when are the v-translate strings sent to Transifex ?

src/views/document/RouteView.vue
brunobesson: It's not mandatory, and in general has no consequences except when used with some old tools, but a good practice.

src/views/document/RouteView.vue
```diff
@@ -210,6 +211,15 @@ export default {
 };
 </script>
 
+var historyWorthActivities = [
```
brunobesson, line 214: use const 🙂

brunobesson:
> Question: how and when are the v-translate strings sent to Transifex ?

Once the PR is merged into `master` branch (this would be nonsense otherwise)

src/views/document/RouteView.vue
brunobesson: Ok. Let's try it this way and see if there is room for improvement later on.

src/views/document/RouteView.vue
```diff
@@ -209,17 +201,15 @@ export default {
...
-
-var historyWorthActivities = [
+const historyWorthActivities = [
         'snow_ice_mixed',
         'mountain_climbing',
         'rock_climbing',
         'ice_climbing',
         'via_ferrata',
         'slacklining',
       ];
```
brunobesson: Oh, didn't get the misplace `</script>` tag 🙁 
I assume I wasn't clear enough: the `historyWorthActivities` declaration should happend before the default export, i.e. before line 128.

Also, could you check the formatting and run prettier? The indentation is false.

Other than that, it's ok for me to merge 👍

src/views/document/RouteView.vue
```diff
@@ -77,7 +77,12 @@
 
         <div class="box">
           <markdown-section :document="document" :field="fields.summary" />
-          <markdown-section :document="document" :field="fields.route_history" />
+          <markdown-section v-if="locale.route_history" :document="document" :field="fields.route_history" />
+          <div v-else-if="showMissingHistoryBanner" class="notification is-info no-print">
+            <edit-link :document="document" :lang="lang" show-always>
+              History is missing, please provide it if you have information.
```
ataraxido, line 83: Missing translation tag/attribute ?

ataraxido (commented):
Not sure however

src/views/document/RouteView.vue
cbeauchesne: Good catch!

cbeauchesne:
```suggestion
              <span v-translate>
                History is missing, please provide it if you have information.
              <span>
```

cbeauchesne (changes requested):

ataraxido (approved):
Okay for me

cbeauchesne (approved):

cbeauchesne:
@AntoineMarnat You can merge!

AntoineMarnat:
And it's translate in Transifex !

PR #1532, opened 2020-12-08
Update get-fundraiser.js

AntoineMarnat:
Update as comments for fundraisers.

### DIFF
--- a/src/js/get-fundraiser.js
+++ b/src/js/get-fundraiser.js
@@ -4,40 +4,40 @@ const mapping = {
   //   url: ' https://www.helloasso.com/associations/rockclimber/formulaires/1/widget'
   // },
   
-  //Greenspit
-  947587: { //Mollans
+  // Greenspit
+  947587: { // Mollans
     url: 'https://greenspits.com/topo/mollans-sur-ouveze/'
   },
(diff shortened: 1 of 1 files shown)


AntoineMarnat:
@cbeauchesne , should I add all the children waypoints of Presles for VTNO, or the main waypoint is enough ?

src/js/get-fundraiser.js
```diff
@@ -1,20 +1,102 @@
 const mapping = {
-  // 14274: { // France
+  // France
+  // 14274: { // FFME - Rockclimber
   //   url: ' https://www.helloasso.com/associations/rockclimber/formulaires/1/widget'
   // },
+  
+  //Greenspit
```
brunobesson, line 7: please add space after `//`, it eases the reading

brunobesson (approved):
If too many cases are added, then this solution won't last :)

src/js/get-fundraiser.js
AntoineMarnat: OK, I'll do.

We should find a way to transmit to the children ...

cbeauchesne:
@brunobesson 

> If too many cases are added, then this solution won't last :)

Definitly :)

Issue #957, opened 2020-01-01
Incorrect georef of outings


asaunier:
When creating an outing, it may happen that users click on the map by mistake (eg. scrolling the page). This results in adding a georef to the outing that is totally incorrect.

An improvement must be found to avoid this. For instance by moving the map to its own section?

momomaniac:
or asking user confirmation if outing too far from route location ? (Threshold to be determined)

cbeauchesne:
or by not allowing users to change to georef of outings ?

asaunier:
> by not allowing users to change the georef of outings ? 

I don't think that would make sense.
It's exactly the same problem than for routes: if no GPX track is loaded, the default point georef could be not very relevant. Being able to move it to some point (for instance an ice fall etc.) would be nice.

AntoineMarnat:
> or asking user confirmation if outing too far from route location ? (Threshold to be determined)

I think it is the best solution. Threshold is not so easy to determine, in particular for the routes where only the summit is attached (access missing). And for outings associated to more than one route (when location is usefull and important).

Maybe 20km from any waypoint associated to or localisation of the routes ?

AntoineMarnat:
> or asking user confirmation if outing too far from route location ? (Threshold to be determined)

Do we agree on that ? 

Completly incorrect georef of outings is a recurrent issue, and some outing may have not been detected. It would be a good enhancement that reduces the work of moderators.

ataraxido:
I would like to suggest a different approach :

- display associated route search before (above) the map
- when a route is picked, pre-filling the outing georef with the one of the route ( like it's done for the title )
- remove the behavior that makes the georef filled when you click on the map
- add a button 'pick from the map' after (at the right) of the latitude field, on the same line

What do you think ?

AntoineMarnat:
At the moment, after one click on the map to set the georef, you can then draw a gps track. 

Your idea seems good to me, I suggest that we also add a button "draw on the map" next to "upload a gpx file". This way, there is no (mis)behaviour when you click on the map without activating a button.

ataraxido:
Should we ask for a community approval ?

cbeauchesne:
IMO, no, it's not a fundamental change of UX.

ataraxido:
Maybe then the approval of people involved in this issue ?

ataraxido:
Working on it I have a question : isn't this current map behavior a problem when adding a route too ?

cbeauchesne:
[...]

PR #1537, opened 2020-12-09
Georef definition enhancements

ataraxido:
The (technical) solution I eventually implemented in order to avoid lat/lng set on click isn't very pretty, that's the least we can say but after many attempts it's the only one I managed to have working : the code is very tangled here with many dependencies.


AntoineMarnat:
To me, the fact that you can drag the items on the map but it is not taken into account, is conterintuitive, and may lead to more missbehaviour.

brunobesson:
Could you explain more what you have done?
De-coupling the fields and the map seems a problem to me.

I also discovered that the lng and lat fields have not the `step` attribute set to `any` => could we fix it too?

ataraxido:
I did exactly what's described here : https://github.com/c2corg/c2c_ui/issues/957#issuecomment-724992431 and seemed to be okay for everyone :)

brunobesson:
> I did exactly what's described here

I have not followed sorry. I'll try to review your code once this behaviour is approved. Could you flag those involved to test your work?

ataraxido:
Hello there @AntoineMarnat @cbeauchesne

AntoineMarnat:
> I did exactly what's described here : [#957 (comment)](https://github.com/c2corg/c2c_ui/issues/957#issuecomment-724992431) and seemed to be okay for everyone :)

What I understood is that the "pick from map" button would allow to drag the georef on the map (locked map otherwise). 

I do not like the fact that what you see on the map may not be what you save, this is against intuition.

What I understood was:
two buttuns: 
- pick georef (below map next to the lat lng)
- draw on map (above map next to upload gpx)

- If neither is activated, the map is frozen: you cannot move geoloc item or gps track.
- Buttons cannot be both activated at the same time.
- If pick georef is activated, you can drag the geoloc only.
- If draw on map is activated, you can create/edit the gps track only.

I understand it might be tricky to implement.

cbeauchesne:
> To me, the fact that you can drag the items on the map but it is not taken into account, is conterintuitive, and may lead to more missbehaviour.

What about not being able to move the picto at all?

src/views/wiki/edition/utils/document-edition-view-mixin.js
```diff
@@ -168,19 +168,6 @@ export default {
       }
     },
 
-    setGeometryPoint() {
```
cbeauchesne, line 171: Removing this function break the edition using numbers inputs now ?

src/views/wiki/edition/utils/MapInputRow.vue
```diff
@@ -24,16 +24,22 @@
           show-recenter-on
         />
       </div>
-      <div class="column is-6">
+      <div class="column is-5">
         <div class="field">
           <label class="label">{{ $gettext('Longitude') }}</label>
-          <input-simple type="number" postfix="°E" v-model="longitude" @input="setGeometryPoint" />
+          <input-simple type="number" step="any" postfix="°E" v-model="longitude" @input="setGeometryPoint" />
```
cbeauchesne, line 30: In case of outings, we should make this inputs read only

cbeauchesne (changes requested):

src/views/wiki/edition/utils/document-edition-view-mixin.js
ataraxido: What do you mean exactly ? Not sure to understand, got no error and everything seems working fine

src/views/wiki/edition/utils/MapInputRow.vue
ataraxido: So the "Pick" button shouldn't be displayed in that case, right ?

ataraxido:
Seems what I proposed does not arrive at a consensus and in the end it's definitely not clear for me what is expected :/

AntoineMarnat:
> What about not being able to move the picto at all?

Yes, whenever the "pick from map" button is not activated.

asaunier:
Could you rename the PR title as "Main implementation" is not very self-explanatory about what it does (and the description does not help either) :P Thanks!

src/views/wiki/edition/utils/document-edition-view-mixin.js
ataraxido: @cbeauchesne ?

Issue #457, opened 2019-04-04
Allow "not known" value for glacier_gear field on routes


cbeauchesne:
It may need an API modification ? 

Techniccly, it means that null should be interpreted as "not known"

AntoineMarnat:
Should we just display "unknown" whenever activity is Snow&Ice and field is null ?

brunobesson:
> Should we just display "unknown" whenever activity is Snow&Ice and field is null ?

I would say so, if we focus on improving the guidebook.

AntoineMarnat:
At the moment, the code reads:
`<field-view v-if="document.glacier_gear != 'no'" :document="document" :field="fields.glacier_gear" />`

Meaning it is displayed only when "yes", aiming to not spam with useless "no" information.

To this extend, I would suggest to do the same with lift access: display only with yes. What do you think ?


I suggest to be slightly more subtle for theglacier_gear field

Display if the value "no" and activity is 
`const glacier_activities = ['mountain_climbing', 'skitouring', 'snow_ice_mixed', 'snowshoeing'];`

Display "unknown" if value is null and activity is ('mountain_climbing' or) 'snow_ice_mixed'.

What do you think ?

brunobesson:
I would agree with your proposal.

> Meaning it is displayed only when "yes", aiming to not spam with useless "no" information.
Not if there is also the undefined value.

AntoineMarnat:
I tried something, but I need to understand more the <field-view>.

cbeauchesne:
call me if you need some help on this point

brunobesson:
I am not sure what to do with this old issue. There is no route in the guidebook where the value is not set. When creating a new route, the default value is no.

Do we need to modify the behavior so that the value is not set bey default (`not specified` presented on th UI side), maybe only for some activities?

I would favor closing this issue.

AntoineMarnat:
I think we can close this issue and remove my branch.

Le mer. 8 mars 2023 à 10:31, Bruno Besson ***@***.***> a
écrit :

PR #1542, opened 2020-12-12
Lifts and glacier_gears  #457

AntoineMarnat:
I do not know how to put a `<v-translate>` as a value for fields.glacier_gear in `<field-view>`.


ataraxido (approved):
Okay for me

brunobesson (commented):
For the code part, ok for me. As for the logic, I am a bit more reluctant now after some tests to always display that there is no need for glacier gear for every skitouring or snowshoeing route.
Seing this for Chamechaude, I'm skeptical...

Maybe we should only display it if different from no or not set?

AntoineMarnat:
> For the code part, ok for me. As for the logic, I am a bit more reluctant now after some tests to always display that there is no need for glacier gear for every skitouring or snowshoeing route.
> Seing this for Chamechaude, I'm skeptical...
> 
> Maybe we should only display it if different from no or not set?

I agree. Maybe show "no" and "unknown" (with edit link) wether empty for 'mountain_climbing' and 'snow_ice_mixed' ? Or only snow_ice_mixed ?

brunobesson:
One should open a small discussion on the forum 🥇

AntoineMarnat:
> One should open a small discussion on the forum 🥇

Done:
https://forum.camptocamp.org/t/champ-materiel-glacier/271849

AntoineMarnat:
I feel like a buck eejit, I just did some tests and there is no such thing as a _null_ value for the relevant activities ... 

We may want a default value to be _unkown_, as we may have _no_ value by default that do not reflect the truth - a problem if we show them. New value means API, I guess. Sorry for my misunderstanding 👎 .

PR #1710, opened 2021-03-09
Doc quality in toolbox

### LINKED ISSUE #1704: Feature: "quality" attribute is misplaced in view pages
In outings, topos, books, etc. the attribute "quality of the document" is currently placed in the middle of other attributes such as activities, route type, configuration, condition ratings, difficulty ratings etc. See for instance:

Screenshot from 2021-03-07 17-06-06

Screenshot from 2021-03-07 17-05-37

This is quite weird because the "quality of the document" is a metadata (ie. data about the document), not an actual piece of data of the route/outing/book/...
Someone reading an outing is most likely not concerned by a mention of the "quality of the document" in the middle of info about the conditions etc.

It may even be rather confusing: someone was surprised because the page of a beautiful photobook had the mention "quality of the document: average", thinking the document was the book...

I would recommend moving the "quality of the document" attribute to some other place in the view pages.
The most consistent place is probably in the left-hand column of the document, with the "versions" link and the licence: 

Screenshot from 2021-03-07 17-16-14

brunobesson:
Fixes #1704 

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -61,8 +61,6 @@ export default {
 </script>
 
 <style lang="scss">
-@import '@/assets/sass/variables.scss';
-
 $sidemenu-width: 200px;
 $body-height: calc(100vh - #{$navbar-height});
 
--- a/src/assets/sass/main.scss
+++ b/src/assets/sass/main.scss
@@ -1,4 +1,5 @@
 @import './variables.scss';
+@import './mixins.scss';
 @import './scrollbar.scss';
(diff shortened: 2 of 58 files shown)


asaunier (approved):
Tested on https://c2corg.github.io/c2c_ui/doc-quality-in-toolbox/#/
I have only looked at the "move quality to toolbox" commit.
The change looks cool to me :)
Good point to have added the quality picto as well!

PR #1715, opened 2021-03-09
Ensure document is defined

lbesson:
Came accross the issue in the console. I didn't really check when and how this.document was supposed to be defined or not


lbesson:
> Could you point to the document that creates the problem? Document should always be defined IMHO

Unfortunately not

ataraxido (approved):

PR #1718, opened 2021-03-10
Add activities to recent outings in route view

### LINKED ISSUE #1716: Add activity for recent outings list in route view
There may be some simplifcations applicable, e.g. if all outings match the route activities.

See https://forum.camptocamp.org/t/nouvelle-sortie/255138/24

brunobesson:
Fixes #1716 

Discussion here: https://forum.camptocamp.org/t/nouvelle-sortie/255138/24


asaunier:
@brunobesson That's great, thanks :)
Tested for instance at https://c2corg.github.io/c2c_ui/recent-outings-activities/#/routes/45738/fr/pointe-des-mattes-depuis-tres-les-pierres

I see that the activities icons are black => why not rendering them in orange as for the other locations where activities are shown?
OK, GPS and images icons are black but so are they in the "associated routes" section in the example page above.

PR #1743, opened 2021-03-18
Rework how errors are displayed

### LINKED ISSUE #1564: Show an error message when an editing conflict occurs
When editing a document, if someone has saved changes in the meantime, the API refuses to save the pushed changes and returns a 409 error.
The UI does not handle this conflict error : nothing happens, one may press the "save" button again but the changes are not saved and it's not clear why saving does not work.

Expected behaviour: at least show an error message explaining why changes cannot be changed and what the user should do.

brunobesson:
Use toast for all alerts.

Fixes #1564

### DIFF
--- a/src/js/vue-plugins/alert-window.js
+++ b/src/js/vue-plugins/alert-window.js
@@ -1,7 +1,0 @@
-export default function install(Vue) {
-  Object.defineProperty(Vue.prototype, '$alert', {
-    get() {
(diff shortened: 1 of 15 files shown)


lbesson:
what does it look like?

brunobesson:
You can try the master branch on .io and provide any invalid file (might be an image). It's a simple red Toast.

Otherwise, I can make a screenshot early next week.

PR #1748, opened 2021-03-22
Small fixes for YETI

iamvdo:
It seems that moment was removed in January, and I was testing on a branch from december, so I didn't detect that before. So, replaced with date-fns.

Also, `v-translate` on slotted elements used to work well, but seems broken now. Fixed also.

### DIFF
--- a/src/components/yeti/SubPanelTitle.vue
+++ b/src/components/yeti/SubPanelTitle.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="columns is-mobile">
     <div class="column">
-      <h2 class="title is-3">
+      <h2 class="title is-3" v-translate>
         <slot></slot>
       </h2>
     </div>
--- a/src/components/yeti/SubPanelBra.vue
+++ b/src/components/yeti/SubPanelBra.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="yeti-subpanel">
-    <sub-panel-title v-translate>Danger level</sub-panel-title>
+    <sub-panel-title>Danger level</sub-panel-title>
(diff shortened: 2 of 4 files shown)


src/components/yeti/SubPanelTitle.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="columns is-mobile">
     <div class="column">
-      <h2 class="title is-3">
+      <h2 class="title is-3" v-translate>
```
brunobesson, line 4: This cannot be used because key extraction will fail. That why I introduced the change on https://github.com/c2corg/c2c_ui/commit/b3246ff7249fb4a0b3f277b139d53440d9c1887c

brunobesson (changes requested):
OK for the moment => date-fns switch.
However the changes to slots and translations is the thing I fixed so that key extraction could work. Unless I did something strange, it is working the way it is right now.

src/components/yeti/SubPanelTitle.vue
iamvdo: Ha, OK, sorry for that. 

The point is there are notices in the console
```

gettext-plugin.js:209 v-translate must contains only text NodeList [div.column]
```

brunobesson: You're right. I did another fix for this, which requires an extra `<span>` tag. I also included your date fix in https://github.com/c2corg/c2c_ui/commit/41be420233058181e21bab3b6b349d05b6cfc1be.

So closing this PR, I'll probably do a release today or on thursday.

PR #1767, opened 2021-03-26
Toolbox is displayed last on mobile

### LINKED ISSUE #1756: Feature: move down the toolbox on mobile
**Community discussion about this feature (mandatory)**
Can't remember when, it's been a while

**Is your feature request related to a problem? Please describe.**
On mobile, document view is inherited from the desktop view, without any logic. The result is :

1. title
1. map
1. toolbox
1. figures
1. textual descriptions
1. outings, routes
1. forum

Which is not so bad, except for toolbox

**Describe the solution you'd like**
Toolbox between textual descriptions and associated docs, or between associated docs and forum

**Describe alternatives you've considered**
Do nothing

brunobesson:
Kind of hacky, but working...
Bulma's columns cannot be used for this, so rely on some specific classes. The trick is use display: contents to skip a hierarchy level.

I applied it on every view when seems applicable (e.g. not on user profile) and tested print view. Might be worth to do some checks on your side too.

Fixes #1756

### DIFF
--- a/src/assets/sass/document-layout.scss
+++ b/src/assets/sass/document-layout.scss
@@ -1,0 +1,53 @@
+@import './variables.scss';
+
+// a kind-of-hacky layout so that for documents, toolbox is displayed last on mobile
+// cannot be done using bulma's columns and column classes...
+.two-columns-doc {
+  display: flex;
+  margin-left: -0.75rem;
+  margin-right: -0.75rem;
+  margin-top: -0.75rem;
+
+  & > div {
+    margin: 0.75rem;
+
+    &:first-child {
+      max-width: 25%;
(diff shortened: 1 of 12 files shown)


cbeauchesne:
To be honest, I'm not a huge fan of this solution : 

* we reinvent the wheel (bulma column), with these consequence:
   * it asks dedicated CSS code (I mean, more complex than a simple straightforward and self-commented line)
   * we are more likely to have display bugs 
   * another thing to understand for next mainteners
* we can't fine tune where the location are

Even if  the actual solution implemented in articleView is not a pure CSS, I feel it's more clear, simple and flexible. WDYT?

PR #1779, opened 2021-03-31
Use web-share API as default

brunobesson:
When addthis is not enabled and web-share api is supported.


brunobesson:
Fixes #1780

PR #1760, opened 2021-03-23
First draft of outing statistics #390

cbeauchesne:

PR #1789, opened 2021-04-06
Add stacked histograms by ratings on outing stats page

cbeauchesne:


src/views/portals/outings-stats/OutingsStatsPart.vue
```diff
@@ -30,8 +70,51 @@ const getOutingMonth = function (outing) {
...
+  const hueOffset =
+    (reference.indexOf(value) - reference.indexOf(minValue)) /
+    (reference.indexOf(maxValue) - reference.indexOf(minValue));
+
+  // going from [0, 1] to [120, 0] (0 means red, 120 means green)
+  const hue = 120 - 120 * hueOffset;
+
+  return `hsl(${hue}, 75%, 50%)`;
+};
+const getRockRatingColor = function (rating, minRating, maxRating) {
```
brunobesson, line 89: add extra empty line

src/views/portals/outings-stats/OutingsStatsPart.vue
```diff
@@ -16,12 +16,52 @@
...
+
+    <div class="columns result-section">
+      <div class="column">
+        <h2 class="title is-4" v-translate>rock_free_rating</h2>
+        <div ref="rock_free_rating"></div>
+      </div>
+      <div class="column">
+        <h2 class="title is-4" v-translate>global_rating</h2>
+        <div ref="global_rating"></div>
+      </div>
```
brunobesson, line 34: A more general comment: depending on activity selected, some graphs are not pertinent. I would rather hide them when it's the case.

src/views/portals/outings-stats/outings-stats.js
```diff
@@ -2,12 +2,23 @@ import d3 from '@/js/libs/d3';
...
-    this.htmlElement = htmlElement;
+    this._data = data;
+    this._htmlElement = htmlElement;
     this._getX = getX;
     this._xTickLabel = (value) => value;
     this._yTickLabel = (value) => value;
     this._getY = () => 1;
+
+    // set the dimensions and margins of the graph
+    this._margin = { top: 10, right: 30, bottom: 50, left: 40 };
```
brunobesson, line 13: could be a const

brunobesson (approved):
Some global remarks however:

* missing months shouldn't be hidden
image
* some graphs should be hidden depending on the activity filtered

asaunier:
What's the aim of this PR? Where would those stats be used/displayed? Don't hesitate to explain the changes in the description :P

src/views/portals/outings-stats/OutingsStatsPart.vue
cbeauchesne: Actually no, because you can have several activity on one outing. And ratings can be missing, even if there is the relevant activity.

=> We can't show/hide graph based on activity, BUT we can hide graph if there is not data. It'll be the object of another PR :)

src/views/portals/outings-stats/outings-stats.js
cbeauchesne: I also could have done a static property, with an immutable object. But as I don't know how it'll be used later (I bet one day or another we'll make this values variables), I assume the best choice for now is it keep it simple.

cbeauchesne:
@asaunier the purpose of this PR is to add ratings graph on outing stats page. For more info about the outing stats, please have a look on #1760: 

image

PR #1805, opened 2021-04-12
Screen plugin optimization

### LINKED ISSUE #1803: Do not load forum data on mobile
On mobile, forum section is not shown. Do not retrieve related data unless it is shown (would happen only on window resize anyway).

### LINKED ISSUE #1804: Enhance screen plugin performance
Use JS media queries instead of resize event.

brunobesson:
Fixes #1804 
Fixes #1803

### DIFF
--- a/src/js/vue-plugins/screen.js
+++ b/src/js/vue-plugins/screen.js
@@ -9,38 +9,60 @@ export default function install(Vue) {
 
     data() {
       return {
-        width: window.innerWidth,
-        height: window.innerHeight,
+        matchingQueryIndex: -1,
       };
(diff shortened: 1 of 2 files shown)


cbeauchesne (commented):
This change breaks this I think : https://github.com/c2corg/c2c_ui/blob/7c70ad5defec47efd233cfcb60bc7dbaaf4876d2/src/views/SideMenu.vue#L13

brunobesson:
> This change breaks

Indeed, I missed it. I have comitted a fix, will produce another release v7.4.17 for this.

PR #1811, opened 2021-04-13
Debug gettext call

brunobesson:


ataraxido (approved):
Ca serait probablement plus logique de faire l'assertion des lignes 162, 163, 164, 165 et 166 avant le call ligne 158 mais rien d'impactant

brunobesson:
You're right. But in fact, this PR has no vocation to be merged. Sorry you were not included in the discussion, but it was done only to help understand the bug observed on Safari. I built the PR because it triggers the deployment on github pages only.

I'll only keep the replacement of `replaceAll` with `replace` + regex. I still need to investigate why it wasn't polyfilled although Safara 13.3 should be in our target.

PR #1815, opened 2021-04-14
Breakdown by activity on outing stats

cbeauchesne:
The first two graphs shows outings count => I added a breakdown by activity

image

### DIFF
--- a/src/views/portals/outings-stats/OutingsStatsPart.vue
+++ b/src/views/portals/outings-stats/OutingsStatsPart.vue
@@ -85,7 +85,7 @@ const getRedToGreenColor = function (value, reference, minValue, maxValue) {
   // going from [0, 1] to [120, 0] (0 means red, 120 means green)
   const hue = 120 - 120 * hueOffset;
 
-  return `hsl(${hue}, 75%, 50%)`;
+  return `hsl(${hue}, 66%, 50%)`;
 };
 
 const getRockRatingColor = function (rating, minRating, maxRating) {
@@ -106,6 +106,28 @@ const getIceRatingColor = function (rating, minRating, maxRating) {
 
 const getHikingRatingColor = function (rating, minRating, maxRating) {
   return getRedToGreenColor(rating, common.attributes.hiking_ratings, minRating, maxRating);
+};
+
+const ACTIVITY_COLORS = {
+  ice_climbing: 'Aqua',
+  snow_ice_mixed: 'LightSteelBlue',
+
+  paragliding: 'MediumBlue',
+
+  hiking: 'ForestGreen',
+  snowshoeing: 'LightGreen',
+
+  rock_climbing: 'SaddleBrown',
+  mountain_climbing: 'RosyBrown',
+
+  skitouring: 'Red',
+  mountain_biking: 'DeepPink',
+  via_ferrata: 'Black',
+  slacklining: 'Orange',
+};
+
+const getActivityColor = function (activity) {
+  return ACTIVITY_COLORS[activity];
 };
 
 // global rating can't be ordered alphabetically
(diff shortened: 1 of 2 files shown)


asaunier:
Thanks for this really cool feature!

I have a suggestion about the legend block: it would be more consistent to sort it the same way than in the graphs: VF and snowshoeing up, ..., hiking at the bottom.
As for now, the sortings are in the reverse order.

src/views/portals/outings-stats/OutingsStatsPart.vue
```diff
@@ -108,6 +108,26 @@ const getHikingRatingColor = function (rating, minRating, maxRating) {
   return getRedToGreenColor(rating, common.attributes.hiking_ratings, minRating, maxRating);
 };
 
+const getActivityColor = function (activity) {
+  return {
+    ice_climbing: 'Aqua',
+    snow_ice_mixed: 'LightSteelBlue',
```
brunobesson, line 114: Why are you using sometimes uppercase, sometimes lowercase?

By the way, I have a personal feeling that predefined colors are not really colors you should use. They are often too flashy. BUt there is a whole topic for these colors on the forum :)

src/views/portals/outings-stats/OutingsStatsPart.vue
```diff
@@ -108,6 +108,26 @@ const getHikingRatingColor = function (rating, minRating, maxRating) {
   return getRedToGreenColor(rating, common.attributes.hiking_ratings, minRating, maxRating);
 };
 
+const getActivityColor = function (activity) {
+  return {
```
brunobesson, line 112: rather store this in a const in order not to instantiate it at every call.

src/views/portals/outings-stats/outings-stats.js
```diff
@@ -126,35 +126,42 @@ export class Histogram {
...
 export class StackedHistogram extends Histogram {
-  constructor(data, htmlElement, getX, getCategory) {
+  constructor(data, htmlElement, getX, getCategories) {
     super(data, htmlElement, getX);
-    this._getCategory = getCategory;
+    this._getCategories = getCategories;
+    this._getCategoryLabel = (category) => category;
     this._color = () => '#F93';
-    this._categoryComparator = (a, b) => (a < b ? -1 : a > b ? +1 : 0);
+    this._categoryComparator = (a, b) => (a < b ? -1 : a > b ? 1 : 0);
```
brunobesson, line 134: Are you sorting strings? Why do you want to reinvent the wheel? https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

src/views/portals/outings-stats/outings-stats.js
```diff
@@ -126,35 +126,42 @@ export class Histogram {
...
 
     for (const d of this._data) {
       const x = this._getX(d);
       const y = this._getY(d);
-      const category = this._getCategory(d);
-      if (category !== undefined && category !== null) {
-        const key = `${x}_${category}`;
+      let categories = this._getCategories(d);
+
+      if (categories !== undefined && categories !== null) {
```
brunobesson, line 148:
```suggestion
      if (!!categories) {
```

src/views/portals/outings-stats/outings-stats.js
```diff
@@ -126,35 +126,42 @@ export class Histogram {
...
+        for (let category of categories) {
+          if (category !== undefined && category !== null) {
```
brunobesson, line 150:
```suggestion
          if (!!category) {
```

src/views/portals/outings-stats/outings-stats.js
```diff
@@ -178,6 +185,11 @@ export class StackedHistogram extends Histogram {
     this._values = Object.values(values);
   }
 
+  categoryLabel(foo) {
```
brunobesson, line 188: `foo` is not really expressive for the meaning of the parameter

brunobesson (approved):

src/views/portals/outings-stats/OutingsStatsPart.vue
cbeauchesne: > Why are you using sometimes uppercase, sometimes lowercase?

Because sometimes I've handwritten it, sometimes copy-paste it :) I'll make this homogenous

> By the way, I have a personal feeling that predefined colors are not really colors you should use. They are often too flashy. BUt there is a whole topic for these colors on the forum :)

Agree, but I don't have the time to look for the perfect set. We'll start with this set, and let anybody who's willing to try to improve it.

src/views/portals/outings-stats/outings-stats.js
cbeauchesne: So, actually, you're right, we use only string categories for the moment. BUT I used this on purpose: it's more generic as it works with any type thats allow `<` and `>` operators.

src/views/portals/outings-stats/outings-stats.js
cbeauchesne: Agree

PR #1829, opened 2021-04-16
Update polyline color: add border around yellow

brunobesson:


asaunier:
Attach perhaps a screen capture?

brunobesson:
Or many, as there are sevaral map backgrounds

Because IGN layer is not working (lazy to get a valid key), I cheated a little bit.

There are some case when the route follows a somehow orange road. But it's still no worse than the yellow line IMHO.

image
image
image
image
image
image

asaunier:
Thanks :)

The new color looks nice to me, but I wonder if yellow is not more visible?
In the example with the swisstopo baselayer around Grand-Saint-Bernard, the track and the road have very similar colors, don't you think?

lbesson:
Isn't openlayer able to "outline" the lines. 
A long time ago, in another project I think we were doing that. (maybe by simply drawing two lines, the one behind being a little bit larger, in the rendering stuff for openlayers)

lbesson:
Something like https://stackoverflow.com/questions/62512673/openlayers-how-to-add-a-colored-border-around-a-line-features

brunobesson:
Nice proposal! I have made some tests with a yellow line and a grey outline, hich render pretty well (at leat, better than single yellow).

Some updated screenshots:
image
image
image
image
image

ataraxido (changes requested):
Code okay but PR title no longer "valid"

ataraxido (approved):
Code okay but PR title no longer "valid"

PR #1855, opened 2021-04-27
Activity breakdown on height diff graphs

cbeauchesne:
From 

image

To 

image


brunobesson (commented):
How do you handle multi-activities outings in this case? Without breakdown, figures were correct. And now?

(no review on the code yet)

cbeauchesne:
I split the value on each activity. Example => `100 m` on an outing with `A` and `B` activities is splited with `50m` on each.

brunobesson (approved):

PR #1877, opened 2021-05-02
Block FLoC via nginx

lbesson:


brunobesson (commented):
Shouldn't it be done on both backend and frontend? In which case varnish would be more appropriate?
@mfournier what do you think?

lbesson:
this is not very clear in the [spec](https://www.w3.org/TR/permissions-policy-1/), but in my understanding this is for defining a permissions policy on a document, so having the header in nginx should be enough (and maybe we could skip the header for js, images etc)

brunobesson (approved):
Could you open a related issue and associate the next milestone to it please? (for the changelog)

lbesson:
Maybe we could also add `camera=(), microphone=(), midi=()` to the Permissions-Policy header

Note: I checked the website of the guardian, which disabled FLoC, and they indeed only send the header on the "page". Maybe we could be smarter with nginx conf and only serve the header for pages

PR #1884, opened 2021-05-04
Hotfix advertising display

cbeauchesne:


src/js/vue-plugins/screen.js
```diff
@@ -10,6 +10,7 @@ export default function install(Vue) {
     data() {
       return {
         matchingQueryIndex: -1,
+        height: window.innerHeight,
```
brunobesson, line 13: `height` will not be updated, so instead, we should add a media query to listen to height 630px breakpoint. I can try a fast branch on my own.

cbeauchesne: Yes I know :(  I do not have the time to do it the right way now :( It's better than nothing, until users do not resize their windows.

> I can try a fast branch on my own.

Ok.

brunobesson: See #1885

Issue #1876, opened 2021-05-01
Feature: Add "soft mobility" icone in outing tables and in outing lists on route page


loicperrin:
**Community discussion about this feature (mandatory)**
https://forum.camptocamp.org/t/annonces-de-mise-a-jour-de-lui/241120/207?u=loic_p

**Is your feature request related to a problem? Please describe.**
The new "soft mobility" icone should be displayed also in table of outings list and in the outing list displayed on a route page

PR #1922, opened 2021-05-17
Add soft mobility icons

brunobesson:
* in outing lists (table view)
* in recent outings list (route, summit)
Fixes #1865


brunobesson:
Fixes #1876

Issue #819, opened 2019-10-17
Feature: Improve route search on outing creation


cbeauchesne:
**Community discussion about this feature (mandatory)**
https://forum.camptocamp.org/t/priorisation-des-developpements/210123/123?u=charlesb , but precise spec need to be discussed on forum

**Is your feature request related to a problem? Please describe.**
It's actually hard to find a route with the actual tool

**Describe the solution you'd like**
TBD

**Describe alternatives you've considered**
TBD

**Additional context**
This feature is at the top position of list asked by community

PR #1937, opened 2021-05-20
Better outing's edition form

### LINKED ISSUE #957: Incorrect georef of outings

cbeauchesne:
Fixes #957, #1981 and #1980 

image

### DIFF
--- a/src/components/map/OlMap.vue
+++ b/src/components/map/OlMap.vue
@@ -120,7 +120,7 @@ import { FIT } from '@/js/fit/FIT';
 import ol from '@/js/libs/ol';
 
 const DEFAULT_EXTENT = [-400000, 5200000, 1200000, 6000000];
-const DEFAULT_POINT_ZOOM = 12;
+const DEFAULT_POINT_ZOOM = 14;
 const MAX_ZOOM = 19;
 const TRACKING_INITIAL_ZOOM = 13;
 
@@ -188,6 +188,11 @@ export default {
       type: Object,
       default: null,
     },
+
+    initialExtent: {
+      type: Array,
+      default: null,
+    },
   },
 
   data() {
@@ -217,6 +222,11 @@ export default {
 
       // layer for document icons and paths
       documentsLayer: new ol.layer.Vector({
+        source: new ol.source.Vector(),
+      }),
+
+      // layer edited document
+      editionLayer: new ol.layer.Vector({
         source: new ol.source.Vector(),
       }),
 
@@ -290,29 +300,20 @@ export default {
     documents: {
       handler() {
         this.drawDocumentMarkers();
-        this.fitMapToDocuments();
       },
     },
 
-    editedDocument: {
-      handler() {
-        this.drawDocumentMarkers();
-        this.fitMapToDocuments();
+    'editedDocument.condition_rating': 'drawDocumentMarkers',
+    'editedDocument.waypoint_type': 'drawDocumentMarkers',
 
-        // when route is set on outing edition, geom may be updated
-        this.setDrawInteraction();
-      },
-      deep: true, // must look on change inside documents object
-    },
-
-    // if, for any reason, geomtry of edited document is set, center map on it
+    // if, for any reason, geometry of edited document is set, center map on it
     // * new value entered in text inputs
     // * first click
     // * new association to route (geometry is copied from route to outing in this case)
     'editedDocument.geometry.geom': {
       handler(to, from) {
         if (from === null && to !== null) {
-          this.fitMapToDocuments(true);
+          this.setDrawInteraction(); // reset interaction mode
         }
       },
     },
@@ -360,6 +361,7 @@ export default {
         this.imagesLayer, // images icons will be under documents
         this.documentsLayer,
         this.waypointsLayer, // keep waypoint above trace and documents
+        this.editionLayer,
       ],
 
       view: new ol.View({
@@ -376,6 +378,8 @@ export default {
 
     this.map.on('moveend', this.sendBoundsToUrl);
     this.map.on('moveend', this.getProtectionAreas);
+    this.map.on('moveend', this.emitMoveEvent);
+
     if (this.protectionAreasVisible) {
       this.protectionAreasLayers.forEach((layer) => layer.setVisible(true));
     }
@@ -394,6 +398,8 @@ export default {
 
     if (this.urlValue) {
       this.view.fit(this.urlValue, { size: this.map.getSize() });
+    } else if (this.initialExtent) {
+      this.fit(this.initialExtent);
     } else {
       this.fitMapToDocuments(true);
     }
@@ -414,7 +420,7 @@ export default {
 
   methods: {
     setModifyInteractions() {
-      const source = this.documentsLayer.getSource();
+      const source = this.editionLayer.getSource();
       const modify = new ol.interaction.Modify({ source });
 
       this.map.addInteraction(modify);
@@ -422,13 +428,13 @@ export default {
 
       modify.on('modifyend', (event) => {
         for (const feature of event.features.getArray()) {
-          this.updateDocumentGeometryFromFeature(feature);
+          this.setDocumentGeometry(feature.get('geometry'));
         }
       });
     },
 
     setDrawInteraction() {
-      const source = this.documentsLayer.getSource();
+      const source = this.editionLayer.getSource();
 
       if (this.drawInteraction) {
         this.map.removeInteraction(this.drawInteraction);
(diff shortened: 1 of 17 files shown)


brunobesson:
Picture? 🥺

cbeauchesne:
It's very early, I'm not yes sure what's the best way to do this.

cbeauchesne:
FYI, I've started a feedback thread on forum: https://forum.camptocamp.org/t/amelioration-du-formulaire-de-sorties/281231

I'll set this PR. "ready for review" once i've finished with comments form users.

cbeauchesne:
Fixes ##957 and #1980 

Ready for review !

src/views/wiki/edition/OutingEditionView.vue
```diff
@@ -190,18 +237,168 @@ export default {
...
+        this.$nextTick(this.fitMapToDocuments);
+      }
+    },
+
+    changeRouteAssociation(addRoute, route) {
+      if (!addRoute && !this.routeIsAssociated(route.document_id)) {
+        return;
+      } else if (addRoute && this.routeIsAssociated(route.document_id)) {
+        return;
+      }
```
brunobesson, line 369: group in one

src/views/wiki/edition/OutingEditionView.vue
```diff
@@ -190,18 +237,168 @@ export default {
...
+            this.fitMapToDocuments();
+            this.updateRoutes('First route', false);
+          });
+        } else {
+          this.$documentUtils.propagateProperties(this.document, route);
+        }
+      } else {
+        this.document.associations.routes = routes.filter((r) => r.document_id !== route.document_id);
+
+        if (this.document.associations.routes.length === 0) {
```
brunobesson, line 391:
```suggestion
        if (!this.document.associations.routes.length) {
```

brunobesson (commented):
Some comments on ergonomics, still:
* do we want to display the routes geom in map? As there is no interaction, what's the use?
* some more already explicited. but it's a first step, we can build on it

src/views/document/utils/boxes/RecentOutingsBox.vue
```diff
@@ -55,9 +55,28 @@ export default {
...
+    showAddOutingButton() {
+      if (this.documentType === 'route') {
+        return true;
+      }
+
+      if (this.documentType === 'waypoint' && this.document.associations.all_routes.total !== 0) {
+        return true;
+      }
+
+      return false;
```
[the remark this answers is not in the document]
cbeauchesne: For this kind of logic, I feel that a `if` for each case is better, because more explicit. I've added a comment on each case.

src/components/map/OlMap.vue
```diff
@@ -471,23 +481,12 @@ export default {
         'addfeatures',
         function (event) {
           event.features.map(this.setDocumentGeometryFromFeature);
-          this.fitMapToDocuments(true);
```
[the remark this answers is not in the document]
cbeauchesne: It is, with a watcher on `document.geometry.geom_detail` in outingEdition component.

I've cleaned the event logic role here : 

1. Except for the very first display, the map component by itself do not decide to refit the map. The use cases to decide to refit or not are too complicated to be in this component, so it's up to the parent component to decide o refit or not.
2. The only use case when the map component refit itself is only when the user decide to do so.

Though, I forgot to add it on route, when the point is set, but not the trace. thank's !

src/views/wiki/edition/OutingEditionView.vue
```diff
@@ -23,12 +24,56 @@
...
+
+      <div v-if="possibleRoutes && possibleRoutes.length !== 0" class="field">
+        <div v-for="route of possibleRoutes" :key="route.document_id">
+          <input-checkbox :value="routeIsAssociated(route.document_id)" @input="changeRouteAssociation($event, route)">
+            <activities :activities="route.activities" class="is-size-4 has-text-secondary" />
+            <document-title :document="route" />,
+            <document-rating :document="route" />
+          </input-checkbox>
+          <document-link :document="route" target="_blank">
+            <fa-icon icon="link" />
```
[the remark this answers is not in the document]
cbeauchesne: I've added a title.

src/views/wiki/edition/OutingEditionView.vue
```diff
@@ -23,12 +24,56 @@
...
+          {{ $gettext('routes') | uppercaseFirstLetter }}:
+          <marker-helper name="1063027#routes" />
+        </label>
+        <div class="control">
+          <input class="input" type="text" placeholder="Name" v-model="routeTitle" />
+        </div>
+      </div>
+
+      <div v-if="possibleRoutes && possibleRoutes.length !== 0" class="field">
+        <div v-for="route of possibleRoutes" :key="route.document_id">
```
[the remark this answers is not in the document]
cbeauchesne: So do I :)

Let's start if this, and see users' feedback

src/views/wiki/edition/OutingEditionView.vue
cbeauchesne: Done. 

Though, could you explain why the second line is better? I really feel that the first one is much more clear, and if there is a performance issue, it should be negligible for a line executed only once, no?

Actually, if it's a good practice to use boolean operator instead of `===0`, we should consider implementing it in the linter, it'll save some time and energy for PR.

cbeauchesne:
Thank's for the review!

> do we want to display the routes geom in map? As there is no interaction, what's the use?

I've been a little bit lazy (or a gentle way to rephrase: I didn't want to do huge PR :) ) I plan to add an title on those icon, and use them as a link, or a selector, don't know yet the best usage.

brunobesson:
> I plan to add an title on those icon, and use them as a link, or a selector, don't know yet the best usage.

Beware, it's complicated to be able to click on such feature in map and open it in a new tab in a native browser way (i.e. click opens in place, ctrl+click or middle mouse click opens in tab or new window). If you want to force to open in blank target, should be doable though.

If you intend to keep the feature in the map, maybe they should have a specific style, to be easily distinguishable.

src/js/vue-plugins/document-utils.js
```diff
@@ -149,11 +149,11 @@ export default function install(Vue) {
...
           this.propagateRoutePropertiesToOutingProperties(child, parent);
-        } else if (parent.type === 'r' && child.type === 'w') {
-          this.propagateGeomPoint(child, parent);
-        } else if (parent.type === 'x' && child.type === 'r') {
-          this.propagateGeomPoint(child, parent);
-        } else if (parent.type === 'x' && child.type === 'o') {
+        } else if (
+          (parent.type === 'r' && child.type === 'w') ||
+          (parent.type === 'x' && child.type === 'r') ||
+          (parent.type === 'x' && child.type === 'o')
```
brunobesson, line 155: parentheses are useless because of operator priority. (unless prettier does a bad job at splitting?)

src/js/vue-plugins/document-utils.js
```diff
@@ -343,7 +343,11 @@ export default function install(Vue) {
       },
 
       getDocumentsBbox(documents) {
-        // given an array of documents, return the extent containing all these documents
+        documents = documents.filter((document) => document?.geometry?.geom);
```
brunobesson, line 346: `document` can be undefined too? maybe you are using too many question marks there 🙂

src/views/wiki/edition/OutingEditionView.vue
```diff
@@ -57,7 +57,7 @@
             <document-title :document="route" />,
             <document-rating :document="route" />
           </input-checkbox>
-          <document-link :document="route" target="_blank">
+          <document-link :document="route" target="_blank" :title="$gettext('Open in a new tab')">
```
brunobesson, line 60: Depending on the browser configuration, I do think it can be a new tab or a new window (although few people use the latter option, I think)

src/views/wiki/edition/RouteEditionView.vue
```diff
@@ -194,6 +189,14 @@ export default {
       }
     },
 
+    onGeometryUpdate(to, from) {
```
brunobesson, line 192: forgot this earlier, but this method is used in many edition views. Could be put in a mixin?

src/js/vue-plugins/document-utils.js
cbeauchesne: Nope, prettier add them.

cbeauchesne:
@b_b, all good on my side

brunobesson (approved):

brunobesson:
Could we consider it fixes #819 too?

cbeauchesne:
Definitely, good catch!

PR #1985, opened 2021-06-03
Add low quality banner for routes and collaborative articles

### LINKED ISSUE #1984: Feature: Disclaimer for empty and draft documents
The aim of this feature is to  emphasis the fact that camptocamp is a collaborative and contributive wiki. 

The idea, following Wikipedia (see for example [here](https://fr.wikipedia.org/wiki/Aphidol%C3%A8te) ), is to add a disclaimer for documents (waypoints, routes, areas, articles, books, but not outings or images) when the completness/quality is either 'empty' or 'draft'.

The disclaimer should be like:

"This document is a draft/empty, feel free to [edit the document] (link to edit page) to provide information and improve its quality"

brunobesson:
Fixes #1984 

@cbeauchesne with the current i18n processing, there is no better way than the one I used to include link within a sentence? I don't think so, but just in case...

### DIFF
--- a/src/views/document/RouteView.vue
+++ b/src/views/document/RouteView.vue
@@ -75,6 +75,7 @@
         </div>
 
         <div class="box">
+          <low-document-quality-banner v-if="!version" :document="document" />
           <markdown-section :document="document" :field="fields.summary" />
           <markdown-section v-if="locale.route_history" :document="document" :field="fields.route_history" />
           <div v-else-if="showMissingHistoryBanner" class="notification is-info no-print missing-history-banner">
@@ -124,6 +125,7 @@
 </template>
 
 <script>
+import LowDocumentQualityBanner from './utils/LowDocumentQualityBanner';
 import documentViewMixin from './utils/document-view-mixin';
 const historyWorthActivities = [
   'snow_ice_mixed',
@@ -133,7 +135,11 @@ const historyWorthActivities = [
   'via_ferrata',
   'slacklining',
 ];
+
 export default {
+  components: {
+    LowDocumentQualityBanner,
+  },
   mixins: [documentViewMixin],
 
   data() {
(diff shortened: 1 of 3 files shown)


src/views/document/RouteView.vue
```diff
@@ -198,13 +204,17 @@ export default {
 
     showMissingHistoryBanner() {
       const doc = this.document;
+      if (this.version || ['empty', 'draft'].includes(doc.quality)) {
+        // hide notice if we are consulting an old version, or if low quality doc banner is already shown
```
brunobesson, line 208: I don't like the fact that the logic to know whether or not a banner for low quality should be displayed is duplicated here, but didn't find any obvious way before going back to my real work...

cbeauchesne: Why not moving `lowQuality` property from LowDocumentQualityBanner component to here ? 

Line 78 would be : 

```
<low-document-quality-banner v-if="showLowQualityBanner" :document="document" />
```

and line 207 would be : 

```
// do not show the two banners
if(this.showLowQualityBanner) { return false; } 
```

Side note, `showLowQualityBanner` should be false for versioned view.

brunobesson: yeah, thought about that too, but I was reluctant because I assumed the responsibility should be that of the component itself. And we would still duplicate the logic for route and article views.

but maybe it's best nonetheless, and we can assume the decision may be that of the view (in which case we could add the warning to other documents with different criteria?)

brunobesson: > Side note, showLowQualityBanner should be false for versioned view.

This is the case already, isn't it?

cbeauchesne: >  I assumed the responsibility should be that of the component itself.

Vaste question :) Usually, when it comes to be entirely displayed, I see these situations : 

1. the decision's logic can change => it's up to the parent to decide
2. the decision's logic is always the same, and is very complex => set it in the child
3. the decision's logic is simple => Do whatever you want :) 
4. you need the decision's result in the parent =>  do not duplicate, set it in the parent.
5.  you need the decision's result in the parent, and it's complex => mixin


Other solution, (but I do not like it, because it causes complex behaviors) : `if (this.$refs.lowQualityBanner.visible)`

> This is the case already, isn't it?

Nope, or did I missed somethin' ? 

```
    lowQuality() {
      return ['empty', 'draft'].includes(this.document.quality);
    },
```

brunobesson: > Nope, or did I missed somethin' ?
Yes, there is a test on `<low-quality-banner>`tag to check if `version` is truthy. But maybe it should move with the whole logic. I'll fix this according to your comments thanks.

What about i18n?

cbeauchesne: > What about i18n?

We hit the limit of our system. 

You assume the in all our langs, "Feel free to" will be before "edit". I bet one :beer: it's not true. 

A hack would be to make the entire sentence a link:

```
    <edit-link show-always :document="document" :lang="lang">
      <span v-translate>Feel free to edit</span>
    </edit-link>
```

(I removed the `translate-context`, IMO it's useless here)

PR #1995, opened 2021-06-07
Route / outing view hide point geometry when applicable

### LINKED ISSUE #1992: Hide single point geometry on route view
Viewing a route point geometry is kind of disturbing for 'normal' users: there is absolutely no interest in it, on the contrary it adds noise to the map, particularly if there are many associated waypoints, notwithstanding a track. So the idea is to hide point geometry for routes. This geometry should still be visible when editing, of course.

The only negative impact I see is that for those working on fixing and placing such point geometry appropriately, you can't just view the document and see if corrections are needed.

### LINKED ISSUE #1994: Hide outing single point geometry on view when there is a track
When an outing's got a GPS track, displaying the point geometry too is useless and a bit disturbing IMHO.
This point geometry should not be hidden when editing, diffing, or viewing a list of documents.

The only downside to hiding this is that the outing's point geometry is colored according to conditions.

brunobesson:
Fixes #1992 
Fixes #1994

### DIFF
--- a/src/components/map/OlMap.vue
+++ b/src/components/map/OlMap.vue
@@ -129,6 +129,14 @@ const respecterCestProtegerService = new RespecterCestProtegerService();
 
 const isNotVirtual = (waypoint) => waypoint.waypoint_type !== 'virtual';
 
+const showSingleDocumentPointGeometry = (document) => {
+  if (document.type === 'o') {
+    // hide point geometry for outing view if GPS data
+    return !document.geometry?.geom_detail;
+  }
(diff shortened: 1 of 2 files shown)


lbesson (approved):

cbeauchesne (commented):
Ok for the outing part.

But for the route part, it would change the way we use the geoloc, and removing a useful feature of c2c (finding the start of a climbing route), so no-go. 

More detail on the discussion on the issue #1992. TL;DR : we cannot make this change without a discussion on the forum.

PR #2013, opened 2021-06-14
Add russian lang for UI

cbeauchesne:


lbesson:
Did we have requests for having the interface in russian?

cbeauchesne:
As far as I know, there is a contributor who proposed to translate it. @AntoineMarnat may have more info.

Side note 1: the process to add a new lang is described here : https://www.camptocamp.org/articles/1205788/en/help-how-to-add-a-new-language-on-camptocamp
Side note 2: Another contributor started to translate it in hungarian.

brunobesson:
We now have added slovenian (and before that chinese) to interface language.
What's the status of hungarian and russian? Should it be added?
@AntoineMarnat @cbeauchesne

AntoineMarnat:
> We now have added slovenian (and before that chinese) to interface language. What's the status of hungarian and russian? Should it be added? @AntoineMarnat @cbeauchesne

For now, Russian and Hungarian are only on the demo to help the translation.

Issue #383, opened 2017-02-02
Sorting cards in advanced search


stef74:
In list of card need to add possibility to order by ***

An good idea should be to have specific color on order criteria (orange ?)

Need to be more specified.

momomaniac:
implementation in progress: #911

PR #2022, opened 2021-06-17
Use API sorting feature in UI 

cbeauchesne:
Generic component for #383

A first try has be done by #911. I have taken a different approach here : 

* The sort tool works for both display: cards and list
* I voluntary keep it very simple: a unique sort key, possibly reversed
  * It corresponds to 95% (from my famous data scientist, the wet finger :smile: ) of usages.
  * having the possibility of more than one key makes the UI way more complex (sorting by "A then B" is not the same as "B then A"). Implementing this UI is very complicated without bad shortcuts, and makes the UI too complex for the 95% above.
* On the same spirit, I tried to display only the most used fields, even if more fields are possible. 

https://c2corg.github.io/c2c_ui/sort-ui/#/routes?act=rock_climbing&sort=-engagement_rating

image

For non-technical comments : https://forum.camptocamp.org/t/trier-les-documents/282278

### DIFF
--- a/src/js/constants/documentsProperties.json
+++ b/src/js/constants/documentsProperties.json
@@ -399,7 +399,6 @@
         "properties": {
           "url": "dhei",
           "activities": [
-            "hiking",
             "rock_climbing",
             "via_ferrata",
             "snow_ice_mixed",
--- a/src/views/documents/utils/QuerySortDropdown.vue
+++ b/src/views/documents/utils/QuerySortDropdown.vue
@@ -1,0 +1,153 @@
+<template>
+  <dropdown-button>
+    <span slot="button" class="button is-size-7-mobile">
+      <fa-icon :icon="['miscs', 'sort']" />
+      <span class="is-hidden-mobile"> &nbsp;{{ $gettext('Sort by') }} </span>
+      <span>&nbsp;</span>
+      <fa-icon icon="angle-down" aria-hidden="true" />
+    </span>
+
+    <div v-if="complexSort">
+      <div class="dropdown-item is-italic" v-translate>You are using a custom sort</div>
+      <div class="dropdown-item">
+        <button class="button" v-translate @click="clear">Reset</button>
+      </div>
+    </div>
+    <div v-else-if="sortableFields.length">
+      <div class="dropdown-item sort-item" v-for="field of sortableFields" :key="field">
+        <span
+          class="has-cursor-pointer is-nowrap sort-label"
+          :class="{ 'has-text-weight-bold': field == sortField }"
+          @click="toggle(field)"
+        >
+          {{ $gettext(field) }}
+        </span>
+        <span v-if="field == sortField" class="has-cursor-pointer">
+          <fa-icon
+            v-if="!reversed"
+            icon="sort-amount-down-alt"
+            @click="setSort('-' + field)"
+            :title="$gettext('ascending sort')"
+          />
+          <fa-icon
+            v-if="reversed"
+            icon="sort-amount-down"
+            @click="setSort(field)"
+            :title="$gettext('descending sort')"
+          />
+        </span>
+        <div v-else class="no-sort-filler" />
+      </div>
+    </div>
+    <div v-else>
+      <div class="dropdown-item is-italic">
+        <span v-if="documentType === 'waypoint'" v-translate>
+          Please select a waypoint type to enable sort feature
+        </span>
+        <span v-translate> Please select an activity to enable sort feature </span>
+      </div>
+    </div>
+  </dropdown-button>
+</template>
+
+<script>
+import urlMixin from './url-mixin';
+
+import constants from '@/js/constants';
+
+export default {
+  mixins: [urlMixin],
+
+  data() {
+    return {
+      sortField: null,
+      reversed: null,
+      complexSort: false,
+      sortableFields: [],
+    };
+  },
+
+  watch: {
+    $route: {
+      handler: 'updateFromRoute',
+      immediate: true,
+    },
+  },
+
+  methods: {
+    setAsComplexSort() {
+      this.sortField = null;
+      this.reversed = null;
+      this.complexSort = true;
+    },
+
+    updateFromRoute() {
+      const sortableFields = [];
+
+      for (const field of Object.values(constants.objectDefinitions[this.documentType].fields)) {
+        if (field.sortable && this.fieldIsVisible(field)) {
+          sortableFields.push(field.name);
+        }
+      }
+
+      this.sortableFields = sortableFields;
+
+      if (!this.$route.query.sort) {
+        this.sortField = null;
+        this.reversed = null;
+        this.complexSort = false;
+        return;
+      }
+
+      if (this.$route.query.sort.includes(',')) {
+        this.setAsComplexSort();
+        return;
+      }
+
+      let sortField = this.$route.query.sort;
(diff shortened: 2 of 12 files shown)


src/views/documents/utils/QuerySortDropdown.vue
```diff
@@ -0,0 +1,148 @@
+<template>
+  <dropdown-button>
+    <span slot="button" class="button is-small-mobile" :disabled="false">
+      <fa-icon icon="sort" />
+      <span class="is-hidden-mobile"> &nbsp;{{ $gettext('Sort by') }} </span>
+      <!-- <span v-if="category.activeCount != 0"> &nbsp;({{ category.activeCount }}) </span> -->
```
brunobesson, line 6: cleanup please

src/views/documents/utils/QuerySortDropdown.vue
```diff
@@ -0,0 +1,148 @@
+<template>
+  <dropdown-button>
+    <span slot="button" class="button is-small-mobile" :disabled="false">
```
brunobesson, line 3: I don't get the `:disabled="false"?

src/views/documents/utils/QuerySortDropdown.vue
```diff
@@ -0,0 +1,148 @@
...
+      <fa-icon icon="angle-down" aria-hidden="true" />
+    </span>
+
+    <div v-if="complexSort" class="content">
+      <div class="dropdown-item is-italic" v-translate>You are using a custom sort</div>
+      <div class="dropdown-item">
+        <button class="button" v-translate @click="clear">Reset</button>
+      </div>
+    </div>
+    <div v-else-if="sortableFields.length !== 0" class="content">
```
brunobesson, line 17: You know my moto with `!== 0` :)

src/views/documents/utils/QuerySortDropdown.vue
```diff
@@ -0,0 +1,148 @@
...
+        <span v-if="field == sortField" class="has-cursor-pointer">
+          <fa-icon v-if="reversed" icon="sort-amount-up" @click="setSort(field)" />
+          <fa-icon v-if="!reversed" icon="sort-amount-down-alt" @click="setSort('-' + field)" />
+        </span>
+      </div>
+    </div>
+    <div v-else class="content">
+      <div class="dropdown-item is-italic">
+        <span v-if="documentType === 'waypoint'" v-translate> Please set a waypoint type to enable sort feature </span>
+        <span v-translate> Please set an activity to enable sort feature </span>
```
brunobesson, line 35:
```suggestion
        <span v-translate> Please select an activity to enable sort feature </span>
```

src/views/documents/utils/QuerySortDropdown.vue
```diff
@@ -0,0 +1,148 @@
...
+        </span>
+        <span v-if="field == sortField" class="has-cursor-pointer">
+          <fa-icon v-if="reversed" icon="sort-amount-up" @click="setSort(field)" />
+          <fa-icon v-if="!reversed" icon="sort-amount-down-alt" @click="setSort('-' + field)" />
+        </span>
+      </div>
+    </div>
+    <div v-else class="content">
+      <div class="dropdown-item is-italic">
+        <span v-if="documentType === 'waypoint'" v-translate> Please set a waypoint type to enable sort feature </span>
```
brunobesson, line 34: I don't understand this one either

src/views/documents/utils/QuerySortDropdown.vue
```diff
@@ -0,0 +1,148 @@
...
+}
+.sort-label {
+  flex-grow: 1;
+  padding-right: 2em;
+}
+
+@media screen and (max-width: $tablet) {
+  .button.is-small-mobile {
+    border-radius: 2px;
+    font-size: 0.7857rem;
```
brunobesson, line 145: another hard coded magic value... 🙁

brunobesson (commented):
* Button positioning is not perfect: spacing is inconsistent with previous buttons. There is too much space before button, not enough before multi-criteria search input.
* Icons seem reversed to me. Arrow down should mean descending order, arrow up ascending. An alt attribute should be added too.

src/views/documents/utils/QuerySortDropdown.vue
```diff
@@ -0,0 +1,148 @@
+<template>
+  <dropdown-button>
+    <span slot="button" class="button is-small-mobile" :disabled="false">
+      <fa-icon icon="sort" />
```
brunobesson, line 4: I would hide this icon when text is displayed, I find it rather confusing because it could also by used as an indicator for a dropdown input, thus we have it twice, somehow. You understand it on mobile, because all buttons have no text. But on desktop, I thought it was a bug first.

Not sure I'm totally clear?

src/views/documents/utils/QuerySortDropdown.vue
cbeauchesne: On waypoints, fields are visible/hidden, depending the waypoint's type.

src/views/documents/utils/QuerySortDropdown.vue
cbeauchesne: I copy pasted from `query-items`, but I'm the one to blame :D 
https://github.com/c2corg/c2c_ui/commit/3814fd1a4ceb45def6b207c2b254ffb7069f3d2e#diff-536a276a9dca18e2ac7816564a4da972e3efed4a191c1547df3148e2bbbe07d9R282

Fixed, I've used bulma helper.

src/views/documents/utils/QuerySortDropdown.vue
cbeauchesne: You are clear. 

But what is your proposition? 

1. Remove the icon? then we should remove it on other dropdown button, it would be strange to have only one button without icon, no? 
2. find a better icon ? Probably, I've been lazy on that one, just took the font-awesome `sort` icon 😅

brunobesson: 3. Keep the icon on mobile, hide in desktop (text suffice).

Option 2 could be ok too.

cbeauchesne: I've changed the icon :  image

cbeauchesne:
> Button positioning is not perfect: spacing is inconsistent with previous buttons. There is too much space before button, not enough before multi-criteria search input.

Yes, I've seen that. Actually, it's not a regression, the bug is present on prod. Now, I treied to fis it, but I failed :( I really do not understand why there is this bug. It's a simple div-block, with div/inline-flex. But there is extra rom between some of child divs. 

> Icons seem reversed to me. Arrow down should mean descending order, arrow up ascending. An alt attribute should be added too.

Done

src/js/constants/documentsProperties.json
```diff
@@ -399,7 +399,6 @@
         "properties": {
           "url": "dhei",
           "activities": [
-            "hiking",
```
brunobesson, line 402: OK, although it should not be in this PR, and an issue should be created too, to track the change 🙂

brunobesson (approved):
I'll have a quick look at the CSS if I can fix it.

brunobesson:
@cbeauchesne I made a commit on your branch to fix the alignment issue.

src/js/constants/documentsProperties.json
cbeauchesne: #2036

PR #2058, opened 2021-07-04
Use a checkbox to make the map filter more obvious

asaunier:
Current is
Screenshot from 2021-07-04 17-05-39

Suggested change:
Screenshot from 2021-07-04 16-57-50

The color and background of the "Filter on map extent" box could probably be improved but don't have a clue yet.
(Code inspired by the "personal feed checkbox")

### DIFF
--- a/src/components/map/OlMap.vue
+++ b/src/components/map/OlMap.vue
@@ -47,13 +47,19 @@
     </div>
 
     <div v-show="showFilterControl" ref="useMapAsFilter" class="ol-control ol-control-use-map-as-filter">
-      <button
-        @click="filterDocumentsWithMap = !filterDocumentsWithMap"
-        :class="{ 'has-text-success': filterDocumentsWithMap }"
+      <input
+        id="filter-documents-with-map"
+        class="switch is-rtl is-rounded is-info is-small"
+        type="checkbox"
+        v-model="filterDocumentsWithMap"
+      />
+      <label
+        for="filter-documents-with-map"
+        v-translate
+        :title="filterDocumentsWithMap ? $gettext('Map filter on') : $gettext('Map filter off')"
       >
-        <fa-icon icon="search" />
-        <span v-translate>Filter on map extent</span>
-      </button>
+        Filter on map extent
+      </label>
     </div>
 
     <div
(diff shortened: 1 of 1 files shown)


cbeauchesne:
> The color and background of the "Filter on map extent" box could probably be improved

Definitly, it's really hard to read the label as is.

Could you try with the dark blue slightly transparent/white font  used for other items ?

asaunier:
@cbeauchesne Thanks for the feedback :)

> Could you try with the dark blue slightly transparent/white font used for other items ?

I guess you mean for instance the "Legend" button in the Yeti maps?
The dark blue doesn't look very good with the blue sliding checkbox (and makes it not that obvious):
Screenshot from 2021-07-05 16-35-53

Using the usual white background with a bold font seems not that bad (the text would be even more readable if using a less transparent background but is it consistent with the opacity of the other white backgrounds?):
Screenshot from 2021-07-05 16-40-02

brunobesson:
You could try either:
* the light background of control items
image
* the background for attributions
image

brunobesson:
You may also try to colo the button with map dark blue?

cbeauchesne:
Agree that the blue background is ugly. IMO The white one seems to be better, but with less alpha.

Also another graphical suggestion: components of map tens to be slightly smaller than usual. Maybe try to reduce the size of the checkbox?

asaunier:
@brunobesson 
> * the background for attributions
> * You may also try to color the button with map dark blue?

Done by f3109b8 : 
Screenshot from 2021-07-06 19-22-19

@cbeauchesne 
> components of map tens to be slightly smaller than usual. Maybe try to reduce the size of the checkbox?

More or less done (not sure it's the right way and I have played with the values to find something empirically) by a77664c : 
Screenshot from 2021-07-06 19-48-38
Screenshot from 2021-07-06 19-48-56

cbeauchesne:
on Chrome, the dot is not located at the good position: 

image

asaunier:
> Have you tried the is-small class ?

Nope, you're right, it's much easier despite it makes the text smaller as well (perhaps it's OK like this?), 77c6ff2
Screenshot from 2021-07-06 22-07-42

cbeauchesne (approved):

PR #2066, opened 2021-07-07
Fix analytics enabling on startup

### LINKED ISSUE #2062: Fix Google Analytics activation
Undo this line change: https://github.com/c2corg/c2c_ui/pull/1646/files#diff-27653c212e1cfe533e4eb2f7d0d3f89604c9de48a09583b4cbbbcbd08a07da79R35

brunobesson:
Because Gdpr plugin `created` function was called before configuring vue-analytics plugin (don't know exactly why...), if the user acceped ga cookies, it was enabled, then disabled again.
This fix ensures to read configuration from stored cookies to enable or disable plugin, although it requires to directly read cookie information.

Not also that we should someday upgrade vue-analytics plugin to vue-gtag...

Fixes #2062

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -29,21 +29,14 @@ Vue.config.silent = false;
 Vue.use(localStorage); // First, vm.$localStorage property
 
 Vue.use(gdpr);
-
-let enableAnalytics = false;
-try {
-  enableAnalytics = JSON.parse(window.localStorage.getItem('Gdpr.preferences', '{}')).choice?.statistics || false;
-} catch (err) {
-  window.localStorage.removeItem('Gdpr.preferences');
-}
 Vue.use(VueAnalytics, {
   id: config.googleAnalyticsKey,
-  disabled: !enableAnalytics,
-  /*debug: {
-    enabled: true,
-    trace: true,
-    sendHitTask: true,
-  },*/
+  disabled: true,
+  beforeFirstHit() {
+    if (Vue.prototype.$gdpr.gdprValue?.statistics) {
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -28,15 +28,22 @@ Vue.config.silent = false;
 
 Vue.use(localStorage); // First, vm.$localStorage property
 
-// Google analytics
+Vue.use(gdpr);
+
+let enableAnalytics = false;
+try {
+  enableAnalytics = JSON.parse(window.localStorage.getItem('Gdpr.preferences', '{}')).choice?.statistics || false;
```
cbeauchesne, line 35: Going deep in localStorage internals may be error prone.

You can use `beforeFirstHit` hook on option sent to VueAnalytics:

```Vue.use(VueAnalytics, {
  id: config.googleAnalyticsKey,
  disabled: true,
  beforeFirstHit() {
    if (Vue.prototype.$gdpr.gdprValue?.statistics) {
      Vue.prototype.$ga.enable();
    }
  },
```

https://github.com/MatteoGabriele/vue-analytics/blob/master/docs/when-google-analytics-is-loaded.md

(`ready()` is called **after** first GA request, so `beforeFirstHit` is the one to go)

cbeauchesne (changes requested):

src/main.js
brunobesson: That was the way to go, thanks.

cbeauchesne (commented):
Review a little bit too late: it works like a charm!

Thank's for the fix!

PR #2082, opened 2021-07-16
Display raw url when printing external link

lbesson:


brunobesson (approved):
Could please create and associate an issue with `enhancement` label and latest milestone? That's for the generated changelog.

lbesson:
In the browser:
image

Currently, when printed:
image

After the change, when printed:
image

lbesson:
possibly, we could decide to use ellipsis for too long urls (the url becomes unusable, but you have at least the hostname, you know it's an link on the web and you can get to the web version of the page if needed)

cbeauchesne (approved):

PR #2087, opened 2021-07-19
Use default lang when creating document if not set

### LINKED ISSUE #2084: Bug: create document - map does not display
https://forum.camptocamp.org/t/pas-daffichage-de-la-carte-lors-ajout-point-de-passage/283995

brunobesson:
Fixes #2084

### DIFF
--- a/src/js/vue-plugins/router.js
+++ b/src/js/vue-plugins/router.js
@@ -108,7 +108,7 @@ const addDocumentTypeView = function (def, viewComponent, editionComponent) {
   });
 
   routes.push({
-    path: '/' + def.documentType + 's/add/:lang?',
+    path: '/' + def.documentType + 's/add/:lang',
     name: def.documentType + '-add',
     component: editionComponent,
   });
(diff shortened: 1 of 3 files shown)


cbeauchesne (changes requested):
I think it does not fixes the issue. the URL is replaced, but the process is not stopped => it calls twice the function load() on `document-edition-mixin`, but the map still have an incomplete locale for a reason I ignore.

That being said, I do not think it's the good solution, as it's a little bit tricky. 

I'd rather do this : 

line 56 on same file : 

```
    lang() {
      return this.$route.params.lang || this.$language.current;
    },   
```

Or solution 2, make the lang param mandatory.

Solution 1 is feature equivalent, solution 2 seems to be a little bit cleaner. Can't make my mind.

brunobesson:
I thought of solution 1 first of course, but found it not really nice not to explicitly plug the lang into the location. Thus my attempt.

Let's make the lang parameter mandatory, then, it's consistent with other routes.

cbeauchesne (approved):

Issue #2101, opened 2021-07-26
Bug: Nofollow on some links, and <meta name="robots" content="noindex" /> in head


cbeauchesne:
There is a lot of history/version/diff results in google... 

* [ ] whatsnew
* [ ] history pages
* [ ] versioned pages
* [ ] edition pages

PR #2105, opened 2021-07-26
Prevent SE to index some pages 

cbeauchesne:
#2101

### DIFF
--- a/src/components/generics/links/AddLink.vue
+++ b/src/components/generics/links/AddLink.vue
@@ -1,7 +1,7 @@
 <template>
   <!-- We alawys display this link, even when user is not logged
     In this case, editions vues will redirect user to /auth -->
-  <router-link :to="{ name: documentType + '-add', params: { lang: $language.current }, query: query }">
+  <router-link :to="{ name: documentType + '-add', params: { lang: $language.current }, query: query }" rel="nofollow">
     <slot>
       {{ $documentUtils.getCreationTitle(documentType) | uppercaseFirstLetter }}
     </slot>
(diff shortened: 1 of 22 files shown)


src/js/vue-plugins/router.js
```diff
@@ -176,4 +196,41 @@ const router = new Router({
...
+  const nearestWithMeta = to.matched
+    .slice()
+    .reverse()
+    .find((r) => r.meta && r.meta.tags);
+
+  // Remove any stale meta tags from the document using the key attribute we set below.
+  Array.from(document.querySelectorAll('[data-vue-router-controlled]')).map((el) => el.parentNode.removeChild(el));
+
+  // Skip rendering meta tags if there are none.
+  if (!nearestWithMeta) return next();
```
brunobesson, line 214: please always use brackets around ifs (I should set a linter rule)

src/js/vue-plugins/router.js
```diff
@@ -176,4 +196,41 @@ const router = new Router({
   },
 });
 
+router.beforeEach((to, from, next) => {
```
brunobesson, line 199: I wonder if this could somehow be handled through vue-meta which we already use?

brunobesson (approved):

src/js/vue-plugins/router.js
cbeauchesne: Definitly !!! Actually, we use `vue-head`. 

I looked for vue-meta, and as I did not found it, I did this small code (didn't want to add new dep).

Let me update this.

cbeauchesne:
@brunobesson I also fixed a small bug. Go from one waypoint to another one, you will see that meta are added, not updated. There are other issue with the same root cause. See here : https://github.com/ktquez/vue-head#replace-content-the-elements

The solution is simply add `id` on unique meta tags. 

Could you redo a review please ?

brunobesson (approved):

PR #2109, opened 2021-07-28
Display selected elevation profile point on map

lbesson:
This is an early draft that when hovering the elevation profile, displays the corresponding point on the map. If the extent of the view doesn't contain the point, the map is centered on the point.

The first question is whether you would be interested in such a functionality. In general I do like that feature, but the map in our use case is quite small, it's not that great imho.

Then, since I'm not a vue expert, your review on the code would be appreciated :) (this is more like a PoC currently, and probably using the bus event is bad practice?)


asaunier:
> The first question is whether you would be interested in such a functionality.

It would be great, thanks :)
Is there already a URL to test your changes?

lbesson:
hum, I think I only pushed this on my fork, so the branch is not built in https://c2corg.github.io/c2c_ui/

brunobesson:
In the meantime, could you also handle window resizing? 🙂 

(no reviexw yet, but I 👍 this proposal)

lbesson:
https://c2corg.github.io/c2c_ui/elevation-profile-map/#/

asaunier:
@lbesson Thanks!
May be tested for instance at https://c2corg.github.io/c2c_ui/elevation-profile-map/#/outings/1333710/fr/mont-maudit-arete-kuffner (using the prod API) => really cool :)

cbeauchesne:
That's soooooooo cool !

Same remark as the other PR, some margin has been eaten : 

image

I don't understand why though Oo

lbesson:
Some updates:

- fit map to documents when switching fullscreen mode
- changed the style for the marker
image
- in fullscreen, and if we have an elevation map, display it at the bottom of the map
image

Ready for your review @asaunier @brunobesson @cbeauchesne

src/views/document/utils/boxes/ElevationProfile.vue
```diff
@@ -296,38 +296,50 @@ export default {
...
+
       this.line.attr('d', this.mode === 'distance' ? this.dLine : this.tLine);
       this.focush.attr('x2', width);
       this.bubble1.attr('x', (width - this.margin.left - this.margin.right) / 2);
       this.bubble2.attr('x', (width - this.margin.left - this.margin.right) / 2);
       this.container.select('rect.overlay').attr('width', width);
-      this.svg.attr('width', width + this.margin.left + this.margin.right);
+      this.svg
+        .attr('width', graphWidth)
+        // while on fullscreen, the x axis switch is displayed in top right corner
```
brunobesson, line 341:
```suggestion
        // when fullscreen, the x axis switch is displayed in top right corner
```

brunobesson (approved):
Nice to have:
* possibility to reduce de profile in fullscreen mode
* better UI, add some kind of viual separation between profile and map? maybe around profile?

PR #2110, opened 2021-07-28
Handle window resize for elevation map

lbesson:


src/views/document/utils/boxes/ElevationProfile.vue
```diff
@@ -95,12 +96,16 @@ export default {
     if (this.hasData) {
       d3.then(this.createChart);
     }
+
+    this.debouncedOnWindowResize = debounce(this.onWindowResize, 300);
```
brunobesson, line 100: I'd rather use a resizeObserver on the graph element for performance reasons, because it's size may not always change when the window is resized.

brunobesson (commented):
Works fine, although there must be a margin lost on the left?
image

And maybe 300 ms is a little bit too much?

lbesson:
I updated the code, should prevent the margin issue

src/views/document/utils/boxes/ElevationProfile.vue
lbesson: ok done

lbesson:
ready for review

lbesson:
ok, fine

Issue #2124, opened 2021-08-02
Bug: the text of the activities icons when printing is too big


lbesson:
**Describe the bug**
![image](https://user-images.githubusercontent.com/127776/127855700-d866aa44-070d-469e-82bd-5d61b5669516.png)

PR #2130, opened 2021-08-03
Correct font size for icon alternate text when printing

lbesson:
solves #2124


cbeauchesne (approved):
Ok for me.

Little suggestion for next PR : use directly repo on c2c org, it would help a little bit reviews with the PR builds :)

PR #2132, opened 2021-08-03
Clean edit link visibility, small refactors

cbeauchesne:
* Edit link visibility is totally controlled by isEditable mixin
* isEditableMixin does not look if user is logged => edit link is displayed as if user where logged
* lowQualityBanner's visibility is handled outside of the component (don't let the component totally decide to be invisible is a common good practice), and depends on isEditable
* * and some small refactors


src/views/document/RouteView.vue
```diff
@@ -25,7 +26,8 @@
             <div class="column is-4">
               <label-value :label="$gettext('ratings')">
                 <document-rating v-if="$documentUtils.hasRating(document)" :document="document" show-helper />
-                <edit-link v-else :document="document" :lang="$user.lang" />
+                <edit-link v-else-if="isEditable" :document="document" :lang="$user.lang" />
+                <span v-else>...</span>
```
brunobesson, line 30: Is this the best choice? Looks like a *Idontknowwhattodointhatfuckingcase* 🙂 
What about `undefined` (`non définies`) instead?

brunobesson (approved):

Issue #2059, opened 2021-07-05
Technical advice on using OLMap or not


iamvdo:
I'm currently working on YETI again, adding new features to the map. I'm facing technical problems and need some advice.

Until now, the `YetiMap` component uses the `OLMap` one as-is. It gives us advantages (basically, same map on YETI than whole website and no need to write similar code twice), but has drawbacks (comes with unwanted features (like filtering) and it's sometimes hard to add new features and leads to more code).

For example, when added drawing capabilities on the map last year, I needed to first remove drawing feature on it (that draws a point and a line) to add specific ones (only lines). It should have been simpler to not rely on `OLMap` at all (less and clearer code).

Now, I want to add new layers based on components and the same problem arises, should I stick with the `OLMap` component or should I create a independent `YetiMap` one (with some code duplication on it)?

Both solutions can work, but I need your opinions/advice on this, and take the *right* solution beforehand.

Thanks!

brunobesson:
AFAIU, building a new map component may be a better solution. I have already introduced async loading for yeti components, so this will have limited impact on user experience. When things differ too much, there is no meaning in keeping it in sync.
However, maybe you could still use some mixins and / or move common methods to utility files?

cbeauchesne:
Hi @iamvdo,

I agree with @brunobesson. I originally thought that a single components would do the job, but there is too much features dedicated for one usage, and mixing all of them make things complicated.

I even hesitated to make a split for view pages and edit pages during my last PR : #1937 

So feel free to build the component that best fits your needs !

iamvdo:
Thanks for your answers. I will follow that path.

PR #2171, opened 2021-08-25
YETI: Refactor + State Management + New avalanche bulletin layer

iamvdo:
Hello,

Here is a PR for YETI. For the review, don't pick each commit as-is, it's a 2-3 months work, so latter commits sometimes revert some previous code.

I'll try to summarize it a bit, this will give you some focus points. Imho, you can check files inside `map-layers` folder, `YetiMap.vue` and `YetiView.vue` (in `/portals`).

## Refactor

First commits are mostly a refactor of the map component. As advised in #2059, YETI no longer uses the same component `OLMap.vue`, but its own version `YetiMap.vue`. It's mostly copy/paste, without unnecessary features. Then, each YETI layer is a component inside `/map-layers` folder.

## State management

As YETI grows, there is a need for state management tool. Instead of loading a new dependency (like vuex), I created a simplified version of it and store everything in `yetix.js` (see b8c4072). Basically:

- Store a global and persistent state shared accross all components
- All mutations occur within the store
- Each component can retrieve state values (in a computed property)
- Provide an eventbus to emit and listen events across map components

## New avalanche bulletin layer

**That is the feature that PR was initially created for**. We're adding a new layer with avalanche bulletin data for each mountains. It's composed by a polygon (mountain area), an icon (danger between 1 and 5) and an overlay (popup with data).
I needed to add a dependency (@turf/intersect) so that icons are always shown on screen, even when zooming (recalculating position).
Right now (august), there is not much avalanche data, so here is fake data if you want to test. Copy/paste this block of code in line 398 of `AvalancheBulletinsLayer.vue` (or check video below)

<details>
  <summary>See full code</summary><p>

```js
// tmp: generate fake data
if (zone.zone === 'Mercantour') {
  zone.danger = {
    low: null,
    max: null,
  };
} else if (zone.zone === 'Ubaye') {
  zone.danger = {
    low: 1,
    max: 1,
    comment: 'Risque faible.',
  };
  zone.orientations = {
    ne: true,
    e: false,
    se: true,
    s: false,
    sw: true,
    w: false,
    nw: false,
    n: true,
  };
} else if (zone.zone === 'Queyras') {
  zone.danger = {
    low: 1,
    lowEvol: 2,
    max: 2,
    comment: 'On est sur un petit danger, mais quand même sur plusieurs lignes',
  };
  zone.orientations = {
    ne: true,
    e: true,
    se: false,
    s: true,
    sw: false,
    w: false,
    nw: false,
    n: true,
    comment:
      'Oh, un commentaire super long qui ne tient pas sur une seule ligne ! Mais va t-on pouvoir le lire en entier ?',
  };
} else if (zone.zone === 'Devoluy') {
  zone.danger = {
    low: 2,
    high: 3,
    max: 3,
    altitude: 2200,
  };
  zone.orientations = {
    ne: true,
    e: false,
    se: true,
    s: false,
    sw: false,
    w: true,
    nw: true,
    n: true,
    comment: 'Plaques récentes',
  };
} else if (zone.zone === 'Pelvoux') {
  zone.danger = {
    low: 3,
    lowEvol: 2,
    high: 4,
    max: 4,
    altitude: 2550,
  };
  zone.orientations = {
    ne: true,
    e: false,
    se: false,
    s: false,
    sw: false,
    w: true,
    nw: true,
    n: true,
  };
} else if (zone.zone === 'Oisans') {
  zone.danger = {
    low: 3,
    lowEvol: 2,
    high: 4,
    highEvol: 3,
    max: 4,
    altitude: 1800,
  };
} else if (zone.zone === 'Vercors') {
  zone.danger = {
    low: 3,
    high: 4,
    highEvol: 5,
    max: 5,
    altitude: 1200,
  };
}
```
</p></details>

Here is a screenshot of what overlay looks like (sorry for long lines of dummy french text :P)

bra-new

And a small video for an overview of that new layer

https://user-images.githubusercontent.com/1260529/130809596-93dc13bc-507e-4604-9280-410813f3e64b.mp4

Thanks for your time, and don't hesitate if you have questions or recommendations.


src/components/yeti/yetix.js
```diff
@@ -0,0 +1,117 @@
+import axios from 'axios';
```
cbeauchesne, line 1: I may have missed something here? It's seems to me that you recoded some Vue reactivity mechanism? you could have the same result with a simple Vue component, and without any trick no ? 

```js
import Vue from 'vue';

export default new Vue({
    data() {
      return {
        activeTab: 0
      }
    },

    methods: {
      setActiveTab(index) {
        this.activeTab = index;
      }
    }
  });
```

src/views/portals/YetiView.vue
```diff
@@ -202,66 +207,24 @@ export default {
         },
         zoom: {
           simple: this.$gettext('Area too large'),
-          full: this.$gettext('Current map view is too large. Please zoom to level ' + VALID_FORM_DATA.minZoom),
+          full: this.$gettext('Current map view is too large. Please zoom to level ' + this.validMinimumMapZoom),
```
cbeauchesne, line 210: Unfortunately, our translation tools supports only raw string. Well, it's not new from this PR, but we'll have to fix it one day.

cbeauchesne (changes requested):
Pretty dense commit :) 

I tried it locally, it seems to be ok.

Though, `npm ci` fails : 

```
> npm ci
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR!
npm ERR!
npm ERR! Missing: @turf/intersect@^6.5.0
npm ERR!

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/charles.debeauchesne/.npm/_logs/2021-09-03T19_39_13_460Z-debug.log
```

Simply execute `npm i @turf/intersect` => it'll update `package-lock.json`

src/components/yeti/yetix.js
iamvdo: Well, I did not recoded Vue reactivity, but used `Vue.observable` (vue's own reactivity system). There is no real difference between what I used and your solution, except that I started with Vuex in mind so ended up with this code (separation between state, mutations and actions). One advantage is that each component can import only what needed, and you can easily find which components will mutate state.

Another point to have in mind is that Yeti now have a persistent state, and if a user go to `/yeti`, then in another page and come back to yeti, the state does not change. To keep things simple for now, I created a `defaultState`, copied it without reference (`JSON.parse(JSON.stringify(obj))`) to pass it to Vue reactivity system. The `setDefaultState` method reverts state back, every time YetiView is mounted. So the same kind of things should be done with your solution.

But I can go this way if you think it's simpler/better! Let me know.

src/views/portals/YetiView.vue
iamvdo: Nice catch. Fixed!

src/components/yeti/yetix.js
cbeauchesne: Sorry, I missed your answer...

> Another point to have in mind is that Yeti now have a persistent state

I may be wrong, but with the old-fashion solution, it's also like that: there is only one `activeTab` value in the entire WPA.

Well, to be honest, I'm definitly for a more solid solution using Vuex for the UI, **but** it would be a entire project. I don't feel that mixing both way of doing is the best solution, it would end with two design choice to understand for new contributors.

That being said, it's not a big deal, as it's used on a very particular part of the UI (I would have a stronger opinion for a plugin used everywhere :) ). So you have my opinion, I let you choose what you want, and I'll commit to your decision as if it would be mine :)

cbeauchesne:
The publication to github page is not working (security reason). 

If it's ok for you, you can merge @iamvdo

cbeauchesne (approved):

iamvdo:
@cbeauchesne Ok for me. I've just rebased the branch on top of master and force pushed.
Also, I added a small modification for Yetix, the state is now read-only. And I added comments.

iamvdo:
Hello. Any news on this?

FYI, I will send another (small) PR soon to handle a future server change. And another one adding a new tool to simplify a loaded route.

It would be really great if all 3 PRs were validated and merged by the end of november.
Thanks!

brunobesson:
@iamvdo could you rebase this branch properly so it can be merged?
the same for #2340

iamvdo:
@brunobesson Hello. I rebased and forced push all 3 PRs. (I don't know why every commit is mentioned in each branch, thus leading to way too many commits...), Maybe because each PR is based on previous one. 

I hope it will be easy to merge. I'm here to help if needed!

PR #2230, opened 2021-09-21
Fine tune events on CI

cbeauchesne:
* Add a workflow_dispatch on changelog
* Add CI on PR (it will be run against the merged version of the code, that's way more relevant
* Remove CI on all pushes (overkill, and add some noise on PR), but keep it at pushes on `master`

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,6 +1,7 @@
 name: Generate changelog
 
 on:
+  workflow_dispatch:
   release:
     types: [published]
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,11 @@
 name: Continuous integration
 
 on:
-  push:
+  pull_request:
     branches:
       - '**'
+  push:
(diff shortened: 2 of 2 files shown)


brunobesson:
OK for allowing to run changelog action manually (I thought it was already the case)

I am not sure to understand your goal on the PR / push side however, what do you intend to do or avoid exactly? Is your goal to avoid running CI on branch pushs if they are not PR branches yet? Then, fine for me.

cbeauchesne:
> Is your goal to avoid running CI on branch pushs if they are not PR branches yet? Then, fine for me.

It'll be a side effect (a positive side effect, agree).

Actually, the thing that annoy me was  that the CI on 'pushes' run on the commit, where the CI on 'pull_requests' runs on the theoretical merged commit, the one who would have be done if you have merged your branch on master.

Sometimes, you may have a commit that is clean, but fails once it's merged.

brunobesson:
I hadn't thought of it, quite nice!

brunobesson (approved):

PR #2232, opened 2021-09-22
Elevation profile enhancements

lbesson:
- Add a button for toggling elevation profile while in profile mode
- Correctly display scales legend (was invisible before)
- Tweak the number of ticks for x axis, to avoid unreadable scale while the elevation profile's width is small

### DIFF
--- a/src/views/document/utils/boxes/ElevationProfile.vue
+++ b/src/views/document/utils/boxes/ElevationProfile.vue
@@ -25,9 +25,13 @@
 
     <div ref="graph" class="elevation-profile-chart" />
 
-    <div class="elevation-profile-fullscreen-toggle" @click="toggleFullScreenProfile">
+    <button
+      :title="$gettext('Toggle elevation profile')"
+      class="elevation-profile-fullscreen-toggle"
+      @click="toggleFullScreenProfile"
+    >
       <fa-icon :icon="toggleIcon" />
-    </div>
+    </button>
   </div>
 </template>
 
@@ -569,6 +573,8 @@ $C2C-orange: red;
     background-color: white;
     cursor: pointer;
     border-radius: 8px 8px 0 0;
+    border: 0;
+    padding: 0;
     box-shadow: $shadow;
   }
 
--- a/src/views/document/utils/boxes/ElevationProfile.vue
+++ b/src/views/document/utils/boxes/ElevationProfile.vue
@@ -474,7 +474,7 @@ export default {
       this.line.transition().duration(1000).attr('d', nLine);
 
       d3.select('.x.axis').call(axis);
-      d3.select('.x.axis.legend').text(legend);
+      d3.select('.x.axis .legend').text(legend);
     },
 
     emitCursorEndEvent() {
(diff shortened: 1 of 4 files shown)


lbesson:
@brunobesson @cbeauchesne I'm not so sure if I did use the font awesome icons properly, since it is used differently elsewhere in the code, nut I think this way of doing is fine?

lbesson:
## Toggle button
Close
image
Open
image

## Legend
image

## Ticks
image
image
image

lbesson:
Added a commit for correcting a bug where the x axis legend was not updated properly

src/views/document/utils/boxes/ElevationProfile.vue
```diff
@@ -24,10 +24,17 @@
...
+
+    <div class="elevation-profile-fullscreen-toggle" @click="toggleFullScreenProfile">
+      <fa-icon :icon="toggleIcon" />
+    </div>
   </div>
 </template>
 
 <script>
+import { faChevronDown } from '@fortawesome/free-solid-svg-icons/faChevronDown';
+import { faChevronUp } from '@fortawesome/free-solid-svg-icons/faChevronUp';
```
brunobesson, line 36: This is @cbeauchesne's work, but the imports are grouped into `font-awesome-config.js`. Don't know the real impact on bundle size & all.

src/views/document/utils/boxes/ElevationProfile.vue
```diff
@@ -24,10 +24,17 @@
     </form>
 
     <div ref="graph" class="elevation-profile-chart" />
+
+    <div class="elevation-profile-fullscreen-toggle" @click="toggleFullScreenProfile">
+      <fa-icon :icon="toggleIcon" />
+    </div>
```
brunobesson, line 30: Add `title` attribute to help users understanding the goal of the button?

brunobesson (approved):
I wonder if we could act on SVG text rendering: it feels no antialiased currently?

src/views/document/utils/boxes/ElevationProfile.vue
lbesson: see https://github.com/c2corg/c2c_ui/pull/2232/commits/e873c860dc7a2905e3044f9892ad8d816b566d18

src/views/document/utils/boxes/ElevationProfile.vue
cbeauchesne: All icons are globally imported and available everywhere. I'm actually surprised it works, because icon prop of `<icon>` should be a string or an array:

* moves imports lines to `src/js/vue-plugins/font-awsome-config.js`
* add the relevant line in this files (see after the import, the how-to is here: https://github.com/c2corg/c2c_ui/wiki/How-to-add-a-new-font-awesome-icon) 
*  this for the function `toggleIcon` : 


```
    toggleIcon() {
        return this.fullScreenHidden ? "chevron-up" : "chevron-down";
```

And IMO, you should not use a function, simply use `<fa-icon :icon="fullScreenHidden ? 'chevron-up' : 'chevron-down'" />` in the template :) 

@brunobesson  even it's never free, FA icons are around few hundred of bytes before compression, it's affordable.

lbesson: It works because it's in the docs :)

Issue #2236, opened 2021-09-23
Bug: avatar updates (via forum profile) are not propagated to c2c_ui


lbesson:
**Describe the bug**
https://forum.camptocamp.org/t/changer-la-photo-de-profil-sur-les-pages-de-sortie/287481/11

PR #2237, opened 2021-09-23
Use most recent avatar version

lbesson:
solves #2236


brunobesson (approved):
I cannot approve the PR without understanding why changing the URL will allow a non-cached avatar image. Care to explain?

But of course I believe in you.

lbesson:
Ok I did some tests by requesting other sizes, and it is indeed a question of cache, not specifically a matter of url.
But who is caching this?

PR #2240, opened 2021-09-27
Print a list of documents in PDF

panglesd:
Hello,

Here is a proposition to add an option to print a pdf containing the list of routes of a research.
For instance, one can find the list of routes located in an area with a suitable difficulty level, then click on the new `Print` button, which will generate a pdf to be saved on a phone or printed.

To do this, I needed a few changes on the organization for document views. Indeed, I needed to display, in a single view, multiple documents, so each document needed to be a component and not a view. I therefore separated the "document api loading" from the displaying of the document. Here is a list of the changes:
- Each AreaView, WaypointView, ... are now viewed as components. `document-view-mixin.js` reflects this. They now take a `promise`, `documentId`, `documentType`, `fields` and `expectedLang` as props instead of finding/loading it themselves
- There is a new view, `DocumentView`, in charge of choosing the right component (AreaView, WaypointView, ...) to load with the right props (`promise`, `documentId`, ...). The route mechanism for document is updated.
- There is a new view, `PrintDocumentList`, which loads all documents, and when everything is resolved, call `window.print()`. The routes for this new view is `/print/{article, waypoint, ...}`
- A button is added to go to the PrintDocumentList from a search, with the same filter.

I hope that this is the right solution to implement this feature! Let me know what you think, I can adapt it to the maintainer's taste. And thanks for this amazing open source project :)

(There are many other natural improvements, such as adding a table of content, but I leave them for another PR.)

### DIFF
--- a/src/js/vue-plugins/router.js
+++ b/src/js/vue-plugins/router.js
@@ -14,6 +14,7 @@ import ProfileView from '@/views/document/ProfileView';
 import RouteView from '@/views/document/RouteView';
 import WaypointView from '@/views/document/WaypointView';
 import XreportView from '@/views/document/XreportView';
+import DocumentsPrintingView from '@/views/documents/DocumentsPrintingView';
 import DashboardView from '@/views/portals/DashboardView';
 import FeedView from '@/views/portals/FeedView';
 import SophiePictureContestView from '@/views/portals/SophiePictureContestView';
@@ -81,6 +82,12 @@ const addDocumentTypeView = function (def, viewComponent, editionComponent) {
     path: '/' + def.documentType + 's',
     name: def.documentType + 's',
     component: DocumentsView,
+  });
+
+  routes.push({
+    path: '/print/' + def.documentType + 's',
+    name: def.documentType + 's-print',
+    component: DocumentsPrintingView,
   });
 
   routes.push({
--- a/src/js/vue-plugins/router.js
+++ b/src/js/vue-plugins/router.js
@@ -85,7 +85,7 @@ const addDocumentTypeView = function (def, viewComponent, editionComponent) {
   });
 
   routes.push({
-    path: '/print/' + def.documentType + 's',
+    path: '/' + def.documentType + 's/print',
     name: def.documentType + 's-print',
     component: DocumentsPrintingView,
   });
--- a/src/views/document/utils/document-view-mixin.js
+++ b/src/views/document/utils/document-view-mixin.js
@@ -95,11 +95,15 @@ export default {
      * properties that are deducted from URL
      */
     documentId() {
-      if (this.isDraftView || this.isPrintingView) return this.draft.document_id;
+      if (this.isDraftView || this.isPrintingView) {
+        return this.draft.document_id;
+      }
       return parseInt(this.$route.params.id, 10);
     },
     documentType() {
-      if (this.isPrintingView) return this.$route.name.split('s-')[0];
+      if (this.isPrintingView) {
+        return this.$route.name.split('s-')[0];
+      }
       return this.$route.name.split('-')[0];
     },
     fields() {
--- a/src/views/document/utils/document-view-mixin.js
+++ b/src/views/document/utils/document-view-mixin.js
@@ -95,9 +95,11 @@ export default {
      * properties that are deducted from URL
      */
     documentId() {
+      if (this.isDraftView || this.isPrintingView) return this.draft.document_id;
       return parseInt(this.$route.params.id, 10);
     },
     documentType() {
+      if (this.isPrintingView) return this.$route.name.split('s-')[0];
       return this.$route.name.split('-')[0];
     },
     fields() {
@@ -174,7 +176,7 @@ export default {
             };
             this.$emit('updateHead');
           });
-      } else if (this.isDraftView) {
+      } else if (this.isDraftView || this.isPrintingView) {
         this.promise = {};
 
         this.$imageViewer.clear();
--- a/src/views/documents/DocumentsPrintingView.vue
+++ b/src/views/documents/DocumentsPrintingView.vue
@@ -1,13 +1,22 @@
 <template>
   <div class="section documents-view">
     <html-header :title="getDocumentTypeTitle(documentType)" />
+    <printing-options
+      :documents="documents"
+      :promises="promises"
+      @set-page-breaks="pageBreaks = $event"
+      @set-image-printing="includeImages = $event"
+      @set-summary-visibility="summaryVisibility = $event"
+    />
     <printing-summary
+      v-if="summaryVisibility"
       :documents="documents"
       :document-type="documentType"
       @summary-click="scrollTo($event)"
     />
     <div :class="['documents-container', includeImages ? 'print-image' : '']">
       <div v-for="(doc, index) in promises" :key="index" class="column">
+        <div v-if="pageBreaks && (index > 0 || summaryVisibility)" class="page-break" />
         <loading-notification :promise="doc.promise" />
         <component
           v-if="doc.promise.data"
@@ -21,6 +30,7 @@
 </template>
 
 <script>
+import PrintingOptions from './utils/PrintingOptions';
 import PrintingSummary from './utils/PrintingSummary';
 
 import c2c from '@/js/apis/c2c';
@@ -36,7 +46,6 @@ import RouteView from '@/views/document/RouteView';
 import WaypointView from '@/views/document/WaypointView';
 import XreportView from '@/views/document/XreportView';
 
-
 export default {
   name: 'DocumentsPrintingView',
 
@@ -47,6 +56,7 @@ export default {
     ImageView,
     MapView,
     OutingView,
+    PrintingOptions,
     PrintingSummary,
     ProfileView,
     RouteView,
@@ -58,6 +68,9 @@ export default {
     return {
       promise: null,
       promises: [],
+      pageBreaks: false,
+      includeImages: false,
+      summaryVisibility: true,
     };
   },
 
@@ -114,3 +127,9 @@ export default {
   },
 };
 </script>
+
+<style scoped>
+.page-break {
+  page-break-after: always;
+}
+</style>
--- a/src/views/documents/DocumentsPrintingView.vue
+++ b/src/views/documents/DocumentsPrintingView.vue
@@ -4,6 +4,7 @@
     <printing-summary
       :documents="documents"
       :document-type="documentType"
+      @summary-click="scrollTo($event)"
     />
     <div :class="['documents-container', includeImages ? 'print-image' : '']">
       <div v-for="(doc, index) in promises" :key="index" class="column">
@@ -97,6 +98,17 @@ export default {
 
     getDocumentTypeTitle(documentType) {
       return this.$gettext(documentType + 's');
+    },
+
+    scrollTo(name) {
+      const el = document.querySelector("#toc-entry-"+name);
+      if (el) {
+        const docEl = document.documentElement;
+        const docRect = docEl.getBoundingClientRect();
+        const elRect = el.getBoundingClientRect();
+        const y = elRect.top - docRect.top;
+        window.scrollTo(0, y - 50); // navbar height ...
+      }
     },
 
   },
--- a/src/views/documents/DocumentsPrintingView.vue
+++ b/src/views/documents/DocumentsPrintingView.vue
@@ -93,18 +93,16 @@ export default {
   methods: {
     load() {
       this.promise = c2c[this.documentType].getAll(this.$route.query).then(() => {
-        this.promises = this.promise.data.documents.map((doc) => {
-          return {
-            promise: c2c[this.documentType].getCooked(
-              doc.document_id,
-              this.$route.params.lang ?? this.$language.current
-            ),
-            documentId: doc.document_id,
-            expectedLang: this.$route.params.lang ?? this.$language.current,
-            documentType: this.documentType,
-            version: null,
-            fields: constants.objectDefinitions[this.documentType].fields,
-          };
+        this.promises = this.promise.data.documents.map((doc) => ({
+          promise: c2c[this.documentType].getCooked(
+            doc.document_id,
+            this.$route.params.lang ?? this.$language.current
+          ),
+          documentId: doc.document_id,
+          expectedLang: this.$route.params.lang ?? this.$language.current,
+          documentType: this.documentType,
+          version: null,
+          fields: constants.objectDefinitions[this.documentType].fields,
         });
       });
     },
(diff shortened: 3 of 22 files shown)


cbeauchesne:
Hi @panglesd ,

Thank you very much for your PR. I'll do a review asap, ETA Wednesday evening.

src/views/documents/DocumentsPrintingView.vue
```diff
@@ -0,0 +1,102 @@
...
+      <div v-for="(doc, index) in promises" :key="index" class="column">
+        <component
+          :is="documentType + '-view'"
+          :document-type="doc.documentType"
+          :version="doc.version"
+          :promise="doc.promise"
+          :document-id="doc.documentId"
+          :expected-lang="doc.expectedLang"
+          :fields="doc.fields"
+        />
```
cbeauchesne, line 15: Page break in a middle of a topo is painfull, we can use this: 

Add a div : `<div class="page-break" />`, and CSS : 

```css
<style scoped>
.page-break {
  page-break-after: always;
}
</style>
```

src/views/documents/DocumentsPrintingView.vue
```diff
@@ -0,0 +1,102 @@
...
+            ),
+            documentId: doc.document_id,
+            expectedLang: this.$route.params.lang ?? this.$language.current,
+            documentType: this.documentType,
+            version: null,
+            fields: constants.objectDefinitions[this.documentType].fields,
+          };
+        });
+        Promise.all(this.promises.map((doc) => doc.promise)).then(() => {
+          Vue.nextTick(function () {
```
cbeauchesne, line 90:
```suggestion
          this.$nextTick(function () {
```

cbeauchesne (changes requested):
Well, there is room for improvment, especially on the loading phase (hard to know when it's ready, when the API is slow), but that is definitly a very good start!

Thanks you so much for your work.

I made some minor remarks and sugesstions, please let me know if you have any comment on them.

src/views/documents/utils/PrintButton.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <router-link :to="path" tag="button" class="button is-primary is-size-7-mobile" @click="exportCsv">
```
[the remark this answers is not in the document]
panglesd: It is an unforgivable remnant of a copy-paste. I will remove it!

src/views/documents/utils/QueryItems.vue
```diff
@@ -51,6 +51,7 @@
       />
       <load-user-preferences-button class="is-hidden-tablet query-item-component" />
       <export-csv-button v-if="listMode" class="is-small-mobile"></export-csv-button>
+      <print-button />
```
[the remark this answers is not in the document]
panglesd: I agree with hiding the button on mobile, and I will try to improve the display.

The current behaviour is that all documents displayed in the page are included in the print. By default, 30 documents are displayed per page, but the user can decide to display 50 or 100 documents per page, and in this case they will all be included in the print. The user can also print "page number 2" of a search. Maybe there can be a tooltip "Print the documents of this page" on the button?

src/views/documents/DocumentsPrintingView.vue
panglesd: I had in mind that it was an improvement to be able to print your topos without wasting every time half a page... but if you think it is better to add a page break I'll include your suggestion.

The best (for the future?) would be that pressing the print button open a piece of UI where user can choose different printing options (keeping image or not, page break at the end of a topo...)

panglesd:
Thanks a lot @cbeauchesne for your review!

Yes quite a few improvements needed, but I wanted to check the base is OK before working on the details. I will try to address your comments (including the loading phase) as soon as possible, but this might happen only early next week...

src/views/documents/utils/QueryItems.vue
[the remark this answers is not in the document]
cbeauchesne: Good ergonomic question, I'll ask on forum

src/views/documents/DocumentsPrintingView.vue
cbeauchesne: > The best (for the future?) would be that pressing the print button open a piece of UI where user can choose different printing options (keeping image or not, page break at the end of a topo...)

Indeed. if you have the motivation, it can be the very first option. Actually, I feel that the auto-open of the print page when documents are loaded is a little bit weird. it could be a good way to go: 

* A `div` at the top of the page, with an parameters' set design. It includes : 
  * a tick box with a label `start each document on top of each page`
  * a disabled button with `loading...`
  * when all documents are loaded, the  `loading...` becomes `print`

WDYT ?

cbeauchesne:
> Yes quite a few improvements needed, but I wanted to check the base is OK before working on the details. I will try to address your comments (including the loading phase) as soon as possible, but this might happen only early next week...

The base is perfect ! thank you very much for your help

src/views/documents/utils/QueryItems.vue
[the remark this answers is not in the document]
cbeauchesne: https://forum.camptocamp.org/t/impression-dune-liste-de-documents-question-dergonomie/288077 

If you do not speak french, i'll translate key points.

panglesd:
Unfortunately there does not seem to be an easy way to generate the page numbers in the table of context in pure css...

Do you think I should rename the table of content to "summary", or delete it right away? I think a summary of all the printed routes is still useful, even without the page numbers.

panglesd:
I am considering looking into how translation work in c2c_ui to have a  french version of this PR's feature.
But before that, @cbeauchesne do you see other modifications to be done?

panglesd:
I'm willing to resume working on this PR, but I might have time for that only in approximately one month.

loicperrin:
Hello @panglesd, have you planned to finalize this PR soon? It is very expected by users :-)

panglesd:
Hello @loicperrin!

Seeing that it seems indeed expected by users, and since I had recently a new idea to make it better, I am indeed motivated to resume working on it!

I have not a lot of time though, but I hope to have the PR in a good state in a few week maximum :)

panglesd:
I will undraft this PR when it's ready for another pass of review.

panglesd:
I think that the PR is now in good shape for reviewing.

I removed the first commit (a refactoring to separate displaying documents from fetching the document) to make the code less intrusive.

I rebased and split the PR in more logical commits.

brunobesson:
Thanks a lot, I'll have a look ASAP!

src/js/vue-plugins/router.js
```diff
@@ -83,6 +84,12 @@ const addDocumentTypeView = function (def, viewComponent, editionComponent) {
     component: DocumentsView,
   });
 
+  routes.push({
+    path: '/print/' + def.documentType + 's',
```
brunobesson, line 88: I wonder if reversing the path wouldn't be more consistent?
```suggestion
    path: def.documentType + 's/print',
```

src/views/document/utils/document-view-mixin.js
```diff
@@ -95,9 +95,11 @@ export default {
      * properties that are deducted from URL
      */
     documentId() {
+      if (this.isDraftView || this.isPrintingView) return this.draft.document_id;
```
brunobesson, line 98:
```suggestion
      if (this.isDraftView || this.isPrintingView) {
        return this.draft.document_id;
      }
```

We try to avoid such shortcut notations in the code, for readablitity.

src/views/document/utils/document-view-mixin.js
```diff
@@ -95,9 +95,11 @@ export default {
...
       return parseInt(this.$route.params.id, 10);
     },
     documentType() {
+      if (this.isPrintingView) return this.$route.name.split('s-')[0];
```
brunobesson, line 102:
```suggestion
      if (this.isPrintingView) {
        return this.$route.name.split('s-')[0];
      }
```

src/views/documents/utils/QueryItems.vue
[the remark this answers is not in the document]
brunobesson: This issue is still not fixed, am I wrong?

src/views/documents/utils/PrintingSummary.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+  <div v-if="documents" class="box">
+    <h1 class="title is-1">Summary</h1>
+    <div v-for="(document, index) in documents.documents" :key="index">
+      <span @click="sendSummaryClick(index)">
```
brunobesson, line 5: There is no indication (e.g. cursor) that clicking on a list item will scroll you there. Also, I suggest to us smooth scrolling to help the user undestand he is scrolling whithin the page, othewise, you might think it opened the document itself. See other review comment.
Also, maybe one should have a shortcut to go back to the top of the page?

src/views/documents/utils/PrintingSummary.vue
```diff
@@ -0,0 +1,48 @@
...
+      default: null,
+    },
+    documentType: {
+      type: String,
+      default: 'route',
+    },
+  },
+
+  methods: {
+
```
brunobesson, line 35: useless empy line

src/views/documents/utils/PrintingSummary.vue
```diff
@@ -0,0 +1,48 @@
...
+    sendSummaryClick(index) {
+      this.$emit("summary-click", index);
+    }
+
```
brunobesson, line 39: useless empty line

src/views/documents/DocumentsPrintingView.vue
```diff
@@ -0,0 +1,135 @@
...
+    },
+
+    scrollTo(name) {
+      const el = document.querySelector("#toc-entry-"+name);
+      if (el) {
+        const docEl = document.documentElement;
+        const docRect = docEl.getBoundingClientRect();
+        const elRect = el.getBoundingClientRect();
+        const y = elRect.top - docRect.top;
+        window.scrollTo(0, y - 50); // navbar height ...
```
brunobesson, line 123:
```suggestion
        window.scrollTo({ top: y - 50, behavior: 'smooth' }); // navbar height ...
```

src/views/documents/utils/PrintButton.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <a target="_blank" :href="path">
+    <button class="button is-primary is-size-7-mobile">Print search results</button>
```
brunobesson, line 3: image

Maybe we could replace the text with a print icon on mobile, to save some space?

src/views/documents/utils/PrintingOptions.vue
```diff
@@ -0,0 +1,117 @@
...
+  },
+
+  computed: {
+    offset() {
+      return parseInt(this.$route.query.offset ?? 0);
+    },
+    total() {
+      return this.promises.length;
+    },
+    allLoaded() {
```
brunobesson, line 59: the name is confusing. shouldn't it be `is-loading` or something similar instead?

src/views/documents/DocumentsPrintingView.vue
```diff
@@ -0,0 +1,135 @@
...
+      @set-page-breaks="pageBreaks = $event"
+      @set-image-printing="includeImages = $event"
+      @set-summary-visibility="summaryVisibility = $event"
+    />
+    <printing-summary
+      v-if="summaryVisibility"
+      :documents="documents"
+      :document-type="documentType"
+      @summary-click="scrollTo($event)"
+    />
```
brunobesson, line 16: I wonder if there could be a way to use columns for displaying pagination, options and print button and use less vertical space?
But it could be for another PR.

brunobesson (commented):
Thanks for the work @panglesd it works well!

Some general remarks otherwise: 

* there are still UI issues with the button positioning: 
image
* there seems to be a problem with included images which are sometimes cropped 
image
* it's a bit confusing to have an option `Include images` which has no effect on images included within the description. Maybe t could be handled too?

panglesd:
Thanks a lot for the review!

I'll try to address the comments (I might not have time before this weekend), and I'll tell you when it's done.

brunobesson:
You're welcome. Thanks for the work 👍

panglesd:
I think I addressed most of your comments!

There are two that are still "pending":
- cropped image. I tried with the css rule `break-inside: avoid` but it does not seem to work well... Maybe I did something wrong, I can try again later.
- Handling images in description:
> it's a bit confusing to have an option `Include images` which has no effect on images included within the description. Maybe t could be handled too?

I am not sure if we should have an option for this, as images included inside a description are generally more of interest than images in the image section. If we handle this case, we should at least provide the option of including only those in the description.
I currently renamed the option to `Include image section` to better reflect what it does. If you think we should, I can include an additional option `Include image in description`.

Thanks!

src/views/documents/utils/PrintingOptions.vue
```diff
@@ -0,0 +1,118 @@
...
+      <div class="column printing-option-section">
+        <h2 class="title is-2">Number of documents</h2>
+        <page-selector :documents="documents" page-type="print" />.
+      </div>
+      <div class="column printing-option-section">
+        <h2 class="title is-2">Other options</h2>
+        <ul>
+          <li><input class="print-option-checkbox" v-model="pageBreaks" type="checkbox" />Add page breaks before each document</li>
+          <li><input class="print-option-checkbox" v-model="imagePrinting" type="checkbox" />Include image section</li>
+          <li><input class="print-option-checkbox" v-model="summaryVisibility" checked type="checkbox" />Include a summary</li>
```
brunobesson, line 13:
```suggestion
          <li><label><input class="print-option-checkbox" v-model="pageBreaks" type="checkbox" />Add page breaks before each document</label></li>
          <li><label><input class="print-option-checkbox" v-model="imagePrinting" type="checkbox" />Include image section</label></li>
          <li><label><input class="print-option-checkbox" v-model="summaryVisibility" checked type="checkbox" />Include a summary</label></li>
```

brunobesson:
> cropped image

I spent 2 minutes on it and couldn't find an obvious solution. Le'ts say we can fix this later.

> I am not sure if we should have an option for this, as images included inside a description are generally more of interest than images in the image section.

I was wondering if we should remove them, but it's not that important. Let's keep it as is for now and see users's feedback later on, if any.

I made one last comment so that checkboxes can be used more easily. Other than that, it's fine for me. I'll merge once you tell me it is ok on your side.

panglesd:
While making another testing pass, I noticed that maps would modify the bbox query argument, updating the search.

So for instance searching, `outings/print?qa=draft,great&bbox=-313650,6588359,202523,6873190` was instantly rewritten in `outings/print?qa=draft,great&bbox=-193317,7142602,-152008,7167790`, retriggering the fetch of all routes and possibly modifying the result. Moreover, moving or zooming on a map was also retriggering the search and route fetch!

I fixed this in the last commit. Once this commit is reviewed, it is good to go on my side!

cbeauchesne:
Thanks a lot @panglesd , that's so great !!

PR #2340, opened 2021-11-16
Yeti: preparation for new server

iamvdo:
This is a small PR in preparation for server update.

It adds attributions (received from the request) to the map. Yeti will soon be able to work on many countries, and so, attributions will follow.

The new wps server is more strict regarding parameters, so the generated URL has to be updated too.

PS: this branch is based on top of yeti-add-bra (https://github.com/c2corg/c2c_ui/pull/2171), only 2 last commits are specific here.

Thanks for review!
Vincent


src/views/portals/YetiView.vue
```diff
@@ -382,11 +382,17 @@ export default {
...
-      result += `potentiel_danger=${potentialDanger};`;
-      result += `neige_mouillee=${wetSnow};`;
-      result += `taille_groupe=${groupSize}`;
+      result += `seuil_alti=${braAltiThreshold}`;
+
+      // specific paramaters based on method
+      if (this.method.type === 'mre') {
+        result += `;rdv=${compass}`;
+      }
+      if (this.method.type === 'mrp') {
```
brunobesson, line 391: could use `else if`, or a `switch`.

src/components/yeti/map-layers/YetiLayer.vue
```diff
@@ -164,6 +162,7 @@ export default {
       this.mapLegend.items.forEach((item) => {
         item.color = `rgb(${item.color[0]}, ${item.color[1]}, ${item.color[2]})`;
       });
+      this.layer.getSource().setAttributions('<li>| ' + this.mapLegend.attributions + '</li>');
```
brunobesson, line 165: is the pipe needed? to me, it should rather be a separator. note that there have been some fixes to attributions (but not in yeti map) in recent commits (your branch is outadated)

brunobesson (approved):

brunobesson (approved):

src/components/yeti/map-layers/YetiLayer.vue
iamvdo: Well, I used pipe to visually separate each attributions (as it is for default ones). Is the commit (878afbd46707494) you're talking about? Not sure if already deployed, but CSS is not applying...

src/views/portals/YetiView.vue
iamvdo: Fix

src/components/yeti/map-layers/YetiLayer.vue
iamvdo: OK, I see. Updated code accordingly.

PR #2342, opened 2021-11-17
Update filling color of activities svg

flob38:
Change filling color to match other activities


brunobesson (commented):
Where does the new color come from?
c2c orange is `#f93`...

Maybe the old color is bad too... 🙁

flob38:
> Anyway, didn't dig the subject, but the SVG color is usually replaced through CSS isn't it?

Yes, but someone who would use it for another support (like poster :) ) would have the wrong color.

Ok so we have to change all svg for they are filling with #ffaa45

cbeauchesne:
Confirm, it's #f93: https://forum.camptocamp.org/t/projet-ui-quel-est-le-orange-c2c/238878/10

flob38:
That why we need a summary : https://www.camptocamp.org/articles/1366655/fr/charte-graphique-camptocamp-association

cbeauchesne (approved):

PR #2470, opened 2022-01-14
Dashboard remodeling for home page_v1

flob38:
Test for discussion
https://forum.camptocamp.org/t/look-de-la-page-daccueil/241462/212


flob38:
Out of date since https://github.com/c2corg/c2c_ui/pull/2858

Issue #2484, opened 2022-01-20
Bug: Licence missing on printed document


flob38:
Licence is missing when printing a document

brunobesson:
@flob38 why is it flagged with `need API modification`?

flob38:
Because I wanted to insert version info : Anyway, wouldn't the most relevant value be the datetime of the last of both revisions? :)
But if you want to decorrelate the to things, feel free.

brunobesson:
Version info is another issue: #1660. I'll remove the `need api modification` label then.

Are you willing / planning to work on this issue? If not, I can assign it to me.

flob38:
Ok.
I've no time for this at the moment.

PR #2485, opened 2022-01-20
Print licence on document

flob38:
To solve #2484

### DIFF
--- a/src/views/document/utils/boxes/ToolBox.vue
+++ b/src/views/document/utils/boxes/ToolBox.vue
@@ -1,159 +1,161 @@
 <template>
-  <div class="box no-print">
-    <associated-documents :document="document" />
+  <div class="box">
+    <div class="no-print">
+      <associated-documents :document="document" />
 
-    <div class="has-text-centered" v-if="isNormalView && available_langs.length">
-      <span v-translate>View in other lang</span>
-      <br />
-      <span class="lang-switcher-box-list">
-        <span v-for="lang of available_langs" :key="lang">
-          <document-link :document="document" :lang="lang">
-            {{ $gettext(lang, 'langs') }}
-          </document-link>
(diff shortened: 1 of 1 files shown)


flob38:
I need some time to work on css

flob38:
Hello.
I'm trying to insert document version with the licence.
I've found 3 versions :/
Exemple : http://localhost:8080/routes/260077/fr/buis-les-baronnies-rocher-saint-julien-le-nez
In the props : 
```
document:Object
    activities:Array[1]
    ...
    cooked:Object
        **version:8**
    ...
    locales:Array[1]
        0:Object
            **version:8**
    ...
    rock_types:Array[1]
    route_types:Array[1]
    type:"r"
    **version:5**
```
It seems that the good one is the one in cooked or locales, but why do the third is different from the others ?
And which one (cooked or locales) should I use ?

cbeauchesne:
The `version=5` is for the non-locales data (height, ratings ....). The `version=8` is for the locales data (description, summary...). 

I would say both values are needed. If it's not possible, the locale one (version=8) is more the one to go.

brunobesson:
Anyway, wouldn't the more relevant value be the datetime of the last of both revisions?

flob38:
That's right but it needs a modification of API no ?
Anyway, if you're ok with that, I'll create a revision-box which would indicate max of revision number. It could be upgraded with date when API will be ok.

brunobesson:
Ok with this, if you have checked that both numbers are related, not independant.

brunobesson:
I did a check. Both version numbers (for locale and 'global') are incremented independently. So the version number you provide is not valid, we need to be able to retrieve dates and compare them to state that this is the version of that date.
Despite your work and efforts, to me this is a no go until API is updated 🙁

PR #2488, opened 2022-01-24
Fix async bug on outing edition

cbeauchesne:
The bug : 

1. Go to the big green button "add a new outing" on the top navbar
2. select an activity
3. selection a route in the list

The map is centered on your route (ok), but the list is now missing. It you move the map to get other route in the same extent, it does not update.

I'm not familiar with `async`/`await` in JS, so another pair of eyes is welcome !

### DIFF
--- a/src/js/get-fundraiser.js
+++ b/src/js/get-fundraiser.js
@@ -148,7 +148,6 @@ const mapping = {
     // Western Australia
     url: 'https://www.climberswa.asn.au/cawa/bolting-fund/',
   },
-  
 
   // Others
   14151: {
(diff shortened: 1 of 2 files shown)


lbesson:
damned

https://github.com/c2corg/c2c_ui/pull/2491

cbeauchesne:
haha, concurrents fixes !

Your PR seems more complete than mine, I may have missed some use case. If yes, let move forward with yours

brunobesson (changes requested):
#2491 seems more complete

lbesson:
let's go with the other one if everyone agrees

PR #2532, opened 2022-02-09
Allow to compute map size on fit when map is hidden

### LINKED ISSUE #2523: Bug: The map on "A proximité" search doesn't work on mobile
**Describe the bug**
On mobile, on a topoguide doc page, when we click on the "A proximité" link, we reach the list of the objects near the preceding object.
But when we show the map, the map extend is the whole world, and the list is updated.
Then the purpose of the "A proximité" link is lost (it's like using the default list of this object type : need many manualy zooms to reach the wanted map extend).

**To Reproduce**

1. On a mobile browser, go to https://www.camptocamp.org/waypoints/45018/fr/dome-de-la-coche
2. Click on 'A proximité' in the bottom box : https://www.camptocamp.org/waypoints?wtyp=summit&bbox=658582,5645264,682766,5665264
3. Check that the first items are in the same area.
4. Click on the 'map' picto to switch to the map.
5. See the map extend : whole world.
6. Click on 'card' picto to switch to card list.
6. Check that the first items aren't in the same area.

**Expected behavior**
The map extend should be ~5 km around the preceding object.

**Desktop :**
- Device: mobile Samsung GS7
- OS: Android
- Browser Firefox
- Version 95.2.0

brunobesson:
Fixes #2523 
The downside being that map tiles are downloaded even if the map is not displayed later on.
Maybe we could try to compute the size based one other DOM elements (e.g. the list), but that could break the isolation and all.

What do you think?

### DIFF
--- a/src/views/documents/DocumentsView.vue
+++ b/src/views/documents/DocumentsView.vue
@@ -347,7 +347,6 @@ $cards-gap: 0.25rem;
   .mobile-mode-both {
     .map-container {
       visibility: hidden; // map does not like to be in a display none...
-      height: 0;
     }
   }
 }


lbesson:
Could we somehow have no background layer when the map is not displayed (initially)?

brunobesson:
> Spent some time trying to do this, but I didn't found any easy way, so let's keep this for later improvement 🙂

PR #2677, opened 2022-04-26
New design feed

flob38:
[On forum](https://forum.camptocamp.org/t/look-de-la-page-daccueil/241462/261)

- Optimize card's size for certains resolutions, in order to see more without scrolling
- Working on the presentation of the cards and their content
- Author and age of card informations should be less visible
- After visiting a link, user should be resume where he left his reading
- Do not show waypoints
- Do not show outings done x weeks before (x to be defined)
- Do not show-up when adding photos (timing ?) on outing
- Do not show-up when a new user is associated on outing
- Do not show-up when a second user add photos on outing
- Add possibility to choose between cards and list
- Let user choose the photos to be shown in feed (3 first images from the last downloading)
- Use quality field to show old outings (to be defined)

### DIFF
--- a/src/components/cards/FeedCard.vue
+++ b/src/components/cards/FeedCard.vue
@@ -60,22 +60,12 @@
     </card-row>
 
     <card-row v-if="item.document.areas && item.document.areas.length">
-      <card-region-item :document="item.document" />
-    </card-row>
-
-    <card-row>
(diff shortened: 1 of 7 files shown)


flob38:
First part : 

- Do not show waypoints
- Add possibility to choose between cards and list

PR #2802, opened 2022-07-12
fit video on phone

BellaChloe:
I was in charge to adapt Youtube videos in a phone format. I modified the Markdown.vue to make it happens.

### DIFF
--- a/src/components/generics/Markdown.vue
+++ b/src/components/generics/Markdown.vue
@@ -284,12 +284,13 @@ div[c2c\:role='video'] {
 
   iframe {
     width: 400px;
-    height: 300px;
+    height: 250px;
     max-width: 100%;
-
+    display: flex;
     @media screen and(min-width: $tablet) {
-      width: 420px;
-      height: 315px;
+      width: 500px;
+      height: 350px;
+      display: flex;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


src/components/generics/Markdown.vue
```diff
@@ -321,4 +321,13 @@ div[c2c\:role='video'] {
     }
   }
 }
+
+@media screen and(max-width: 844px) {
```
brunobesson, line 325: Where does this value `884` come from? We use bulma breakpoints to distinguish between phone, tablet, desktop and widescreen (768, 1023, 1215 ,1407, +- 1px depending on comparison made)

brunobesson (dismissed):
Thanks for the PR.

Could you please add some description to your PR? What it intends to do, and how? I can guess, but it would ease the discussion a lot.

src/components/generics/Markdown.vue
BellaChloe: When I put $phone on my doc, it has got an error and I wasn't able to do my internship mission correctly.In school, I learnt to add some value in this case. On the inspector,it was written this value, I wrote this value and the error disappeared.

cbeauchesne (changes requested):
We just met with Chloé. So : 

* fix the error (`phone` variable does not exists)
* Use a simpler way to achieve that
* include the diff in existing CSS descriptors

cbeauchesne (approved):

PR #2812, opened 2022-07-18
Button fav

BellaChloe:
From last week, after that I put my previous doc on the Git, I wanted to try to add a new button on the outing file. This fav button is essential for users who want to add an outing to their profile and want also to prepare a trip and haven't the wifi or mobile network when they will make their trail or their outing the D-Day. To do it, I based myself on View documentation principally and also the forum. It's and idea that I got and that I can modify if it's needed.

### DIFF
--- a/src/components/generics/DocumentField.vue
+++ b/src/components/generics/DocumentField.vue
@@ -2,8 +2,8 @@
   <!-- This a generic component that display field value of a document for any kind of field  -->
   <span>
     <activities v-if="field.name == 'activities'" :activities="document.activities" class="has-text-secondary" /><!--
-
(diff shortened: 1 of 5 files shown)


asaunier:
Hey @BellaChloe 
Could you precise how this tool would be used? Screenshots?
Are you sure people would save an outing to prepare their own? Saving a topo ("route") would make more sense I think.
By the way there is already a favorite routes tool that people may use to mark a topo as "todo", see the star icon in the top right corner of topos (eg. https://www.camptocamp.org/routes/54092/fr/mont-blanc-traversee-domes-de-miage-aiguille-de-bionnassay-mont-blanc) and the "My favorites" menu in the user's dropdown.

PR #2870, opened 2022-08-24
Add base carcass for PWA

cbeauchesne:
Based on the work from @BellaChloe (#2860), here is a basis for a first carcass of a PWA.

The goal here is only to enable page loading of UI base components in offline mode. Testing is not straightforward, as the dev server, by its own nature, disable any cache on browser side. To test it, here is the recipe : 

``` bash
npm run build # will build the production files
cd dist # go to build's folder
python -m SimpleHTTPServer 8000  # a very convenient way to start a simple file server. Python rocks 
```

Now, use Chrome : 

* open a **private** tab (it works in normal mode of course, but you will avoid a lot of side effects caused by this hack)
* Load a page at http://localhost:8000/
* got to dev tools > lighthouse
* untick all but PWA
* Process the report
* All must be green ! 

To double check :

* Go to `Network tab` in dev tools
* Set throttling to `offline`, it will simulate offline mode
* Reload the page => it works

----

As it, it's pretty useless as the content from the API will miss. It will come as a next step!


package.json
```diff
@@ -40,6 +40,7 @@
     "file-saver": "2.0.5",
     "ol": "7.0.0",
     "process": "0.11.10",
+    "register-service-worker": "^1.7.2",
```
brunobesson, line 43: please use fixed version

package.json
```diff
@@ -58,6 +59,7 @@
     "@snyk/protect": "1.991.0",
     "@vue/cli-plugin-babel": "5.0.8",
     "@vue/cli-plugin-eslint": "5.0.8",
+    "@vue/cli-plugin-pwa": "^5.0.8",
```
brunobesson, line 62: please use fixed version

package-lock.json
```diff
@@ -1,13682 +1,8 @@
 {
   "name": "c2c_ui",
   "version": "7.6.10",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
brunobesson, line 4: please use recent node version and keep lockfile v2. Current node LTS is 16 and this is the version targeted in CI.

public/robots.txt
```diff
@@ -1 +1,3 @@
 Sitemap: https://api.camptocamp.org/sitemaps.xml
+User-agent: *
+Disallow:
```
brunobesson, line 3: what's the use here?

brunobesson (commented):
It works, indeed :)
Howerever, what's the purpose of this PR? Do you intend to merge it as is? On my personal side, I have already give it some thought and made little expriments.

Just some thoughts:
* when offline, the app is continuously trying to fetch feed data
* for now, you have a 'base carcass' as you wrote. the minimal goal of an offline c2 app, ignoring all the installable stuff, would be to allow to consult the last documents viewed. And this would somehow involv storing data in a db (indexeddb?). What's your curent goal?

cbeauchesne:
> Howerever, what's the purpose of this PR? Do you intend to merge it as is? 

Only having a base for a PWA, no usable feature is planned for this PR. As it (should) have no side side effect, it was enough for a trainee period. But good catch for the feed, I missed that 🤦.  We can't merge it with that behavior, I'll fix it. 

> What's your curent goal?

Just preparing the field for anyone who want to do the real feature. Speaking of which, I think that "store last document" may not be the good idea, as it will be capped, and you won't be able to rely on it before going offline (let's imagine you consult a lot of topo at the hut after consulting the route you want to do....)

IMO, the feature should look like : 

* a button "Make this document available offline" at the top of any document page
* a button in the profile dropdown "Stored documents" leading to ...
* ... a page  with the list of stored document

But it'll be for another PR.

brunobesson:
> Just preparing the field for anyone who want to do the real feature.

OK, then. What about setuping workbox? What made you choose `register-service-worker` over it? It seems to me it is a more widely used solution? But maybe it depends on the vision we have for c2c as a PWA?

> But it'll be for another PR.

Then at least remove console.log too 🙏

cbeauchesne:
The [first commit](https://github.com/c2corg/c2c_ui/pull/2870/commits/0e4f08f50f1afcefa5b37ba371ac825099df1ac7) is the exact result of the execution of `vue add pwa`. So : 

> What made you choose register-service-worker over it?

Because it was made by the command above, no vision here :D 

Same with `console.log`, but I understand, I will remove them.

PR #2909, opened 2022-09-14
YETI: new layers

iamvdo:
Hi,

This PR adds 2 new layers to YETI app: Nivoses and ROMMA layers

Here is a screenshot
image

It also fixes some bugs and/or improves few things.

Thanks

### DIFF
--- a/src/components/yeti/map-layers/IconLayer.vue
+++ b/src/components/yeti/map-layers/IconLayer.vue
@@ -161,7 +161,7 @@ export default {
 </script>
 
 <style scoped lang="scss">
-@import '@/assets/sass/variables.scss';
+@import '@/assets/sass/variables';
 
 .overlay {
   min-width: 150px;
--- a/src/components/yeti/map-layers/IconLayer.vue
+++ b/src/components/yeti/map-layers/IconLayer.vue
@@ -1,0 +1,201 @@
+<template>
+  <div ref="overlay" class="overlay">
+    <p class="overlay-header" :style="{ background: color }">
+      <slot></slot>
+    </p>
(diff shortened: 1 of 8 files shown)


brunobesson (approved):

brunobesson:
Thanks for the great work 🙂

brunobesson:
Let's fix the remaining isus and merge this @iamvdo ?

src/components/yeti/map-layers/IconLayer.vue
```diff
@@ -0,0 +1,201 @@
...
+        this.$parent.unsetOverlay();
+        this.activeFeature = false;
+      }
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
+@import '@/assets/sass/variables.scss';
```
brunobesson, line 164: You should remove the extension here.

```suggestion
@import '@/assets/sass/variables';
```

iamvdo: OK, but why only this specific file? There are many imports with .scss in them...

brunobesson: I don't think so?
image

It was before, but not since 898c4e43d5f1661161d26e83e16de4b36d9d3a86
Maybe you missed it when rebasing?

iamvdo: OK, so I had to remove from all yeti files then! :)

iamvdo: Oh, I was checking an old branch, my bad. 
I fix it

PR #2930, opened 2022-09-22
Provide health on http for k8s

lbesson:
Au doigt mouillé 😃


brunobesson (approved):
Gros doigt mouillé, but let's try like this.

PR #2937, opened 2022-09-26
Fix #2689

### LINKED ISSUE #2689: Bug: not possible to create an outing on today's date (grey - not possible to select)
**Describe the bug**
When creating anouting, not possible to select the date of today, which is grey. 
refer to : https://forum.camptocamp.org/t/bug-date-sortie/297831/6

**To Reproduce**
Open a navogator sheet
Wait until the day after
try to create an outing
the date of today cannot be selected.

**Expected behavior**
Be able to create an outing on today's date even if the navigator is open since the (or another) day before.

**Screenshots**
no

**Desktop (please complete the following information):**

- Device: PC
- OS: all
- Browser all
- Version all
- 
**Additional context**
Sheet opened since the day or another day before.

Solution : 
The maximum date must be updated at date of the loading of the outing creation and not at the date of sheet opening.

jeanphilippeds:
Hi,

Here is a small PR to fix #2689 by setting the max date when the user clicks on the form field and not at the first mount time.
At the same time, I think I fixed the old min/max criterias that were not taken into account by form-field component.

Have a good day,
JP

### DIFF
--- a/src/views/wiki/edition/OutingEditionView.vue
+++ b/src/views/wiki/edition/OutingEditionView.vue
@@ -10,7 +10,7 @@
           class="is-narrow"
           :document="document"
           :field="fields.date_start"
-          :max="currentDate || (showBothDates ? document.date_end : undefined)"
+          :max="showBothDates ? document.date_end : currentDate"
           @input="handleDates"
           @click.native="setCurrentDate"
         />
--- a/src/views/wiki/edition/OutingEditionView.vue
+++ b/src/views/wiki/edition/OutingEditionView.vue
@@ -10,8 +10,9 @@
           class="is-narrow"
           :document="document"
           :field="fields.date_start"
-          :max="showBothDates ? document.date_end : undefined"
+          :max="currentDate || (showBothDates ? document.date_end : undefined)"
           @input="handleDates"
+          @click.native="setCurrentDate"
         />
         <form-field
           class="is-narrow"
@@ -19,6 +20,8 @@
           :document="document"
           :field="fields.date_end"
           :min="showBothDates ? document.date_start : undefined"
+          :max="currentDate"
+          @click.native="setCurrentDate"
         />
         <div class="column is-narrow">
           <input-checkbox v-model="showBothDates">{{ $gettext('Several days?') }}</input-checkbox>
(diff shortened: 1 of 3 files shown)


src/views/wiki/edition/OutingEditionView.vue
```diff
@@ -396,7 +400,9 @@ export default {
         this.document.date_end = this.document.date_start;
       }
     },
-
+    setCurrentDate() {
+      this.currentDate = new Date().toISOString().substring(0, 10)
```
brunobesson, line 404: This conversion to stick to a date (which you only copied) is a bit ugly IMHO.
We use `date-fns` on c2c_ui, so using `format(new Date(), 'yyyy-MM-dd')` would seem more appropriate IMHO.

(applies also to line 246)

src/views/wiki/edition/OutingEditionView.vue
```diff
@@ -10,15 +10,18 @@
           class="is-narrow"
           :document="document"
           :field="fields.date_start"
-          :max="showBothDates ? document.date_end : undefined"
+          :max="currentDate || (showBothDates ? document.date_end : undefined)"
```
brunobesson, line 13: By doing this, you allow to select 2 dates with the first one being after the last one, because `currentDate` is always set and thus the second part is never evaluated.

Instead, it should be `currentDate` if single date, but the min of `document_date_end` and `currentDate` otherwise.

brunobesson (changes requested):
Thanks for your work. It looks good to me, except the upper limit computation with a start and end dates, which needs to be computed another way.

I also suggest to format the date using `date-fns` rather the current substring splicing to make it more explicit, but it's a matter of taste, maybe.

src/views/wiki/edition/OutingEditionView.vue
jeanphilippeds: Oh I did not know you were using date-fns. Way better indeed :)

src/views/wiki/edition/OutingEditionView.vue
jeanphilippeds: Good catch!
I think a clean condition could be: `showBothDates ? document.date_end : currentDate` as we know that `document.date_end < currentDate`

brunobesson (approved):

PR #2947, opened 2022-10-02
Add support of version masking

asaunier:
API PR at https://github.com/c2corg/v6_api/pull/1393

### DIFF
--- a/src/components/generics/links/MaskLink.vue
+++ b/src/components/generics/links/MaskLink.vue
@@ -1,10 +1,12 @@
 <template>
   <span>
     <span v-if="$user.isModerator" @click="toggleMask">
-      <a v-if="isVersionMasked" v-translate>Unmask version</a>
-      <a v-else v-translate>Mask version</a>
+      <a v-if="version.masked" v-translate>Unmask</a>
+      <a v-else v-translate>Mask</a>
     </span>
-    <span v-if="isVersionMasked" v-translate>This version is masked.</span>
+    <span v-if="version.masked">
+      [<span v-translate>This version is masked.</span>]
+    </span>
   </span>
 </template>
 
@@ -30,21 +32,15 @@ export default {
     },
   },
 
-  computed: {
-    isVersionMasked() {
-      return this.version.masked;
-    }
-  },
-
   methods: {
     async toggleMask() {
       try {
         await (
-          this.isVersionMasked ?
+          this.version.masked ?
           c2c.moderator.unmaskVersion(this.id, this.lang, this.version.version_id) :
           c2c.moderator.maskVersion(this.id, this.lang, this.version.version_id)
         );
-        this.isVersionMasked = !this.isVersionMasked;
+        this.version.masked = !this.version.masked;
       } catch (error) {
         toast({
           message: this.$gettext(`An error occurred, couldn't (un)mask version`),
--- a/src/components/generics/links/MaskLink.vue
+++ b/src/components/generics/links/MaskLink.vue
@@ -1,0 +1,57 @@
+<template>
+  <span>
+    <span v-if="$user.isModerator" @click="toggleMask">
+      <a v-if="isVersionMasked" v-translate>Unmask version</a>
+      <a v-else v-translate>Mask version</a>
+    </span>
+    <span v-if="isVersionMasked" v-translate>This version is masked.</span>
+  </span>
+</template>
+
+<script>
+import { toast } from 'bulma-toast';
+
+import c2c from '@/js/apis/c2c';
+
+export default {
+
+  props: {
+    version: {
+      type: Object,
+      required: true,
+    },
+    lang: {
+      type: String,
+      required: true,
+    },
+    id: {
+      type: Number,
+      required: true,
+    },
+  },
+
+  computed: {
+    isVersionMasked() {
+      return this.version.masked;
+    }
+  },
+
+  methods: {
+    async toggleMask() {
+      try {
+        await (
+          this.isVersionMasked ?
+          c2c.moderator.unmaskVersion(this.id, this.lang, this.version.version_id) :
+          c2c.moderator.maskVersion(this.id, this.lang, this.version.version_id)
+        );
+        this.isVersionMasked = !this.isVersionMasked;
+      } catch (error) {
+        toast({
+          message: this.$gettext(`An error occurred, couldn't (un)mask version`),
+          type: 'is-danger',
+        });
+      }
+    },
+  },
+};
+</script>
--- a/src/components/generics/links/MaskLink.vue
+++ b/src/components/generics/links/MaskLink.vue
@@ -3,9 +3,7 @@
     <span v-if="$user.isModerator" @click="toggleMask">
       <a v-translate>Toggle masking</a>
     </span>
-    <span v-if="version.masked">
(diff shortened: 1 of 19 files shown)


src/js/apis/c2c/ModeratorService.js
```diff
@@ -61,4 +61,20 @@ ModeratorService.prototype.unblockAccount = function (userId) {
   return this.api.post('/users/unblock', { user_id: userId });
 };
 
+/**
+ * Mask/unmask version service
+ */
+ModeratorService.prototype.maskVersion = function ({ documentId, lang, versionId }) {
+  console.log(documentId);
```
brunobesson, line 68: remove console.log

asaunier: Don't worry, the branch is only at its very early stage :P

src/components/generics/links/MaskLink.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <span>
+    <span v-if="$user.isModerator" @click="toggleMask">
```
asaunier, line 3: There is probably something wrong around here because when a moderator clicks on the "Mask" link of an archived version
image

the latter link title remains "Mask" instead of toggling to "Unmask" as expected. See:
image

After reloading the page, the link title is OK though:
image

Same problem when clicking on "Unmask" then.

asaunier: @brunobesson Would you have a clue about this problem? Thanks :)

lbesson (commented):
A bit following the remarks on the API PR:

- normal users shouldn't be able to see a diff between a masked version an an (un)masked one. This should be prevented by the API, but the diff links should also be disabled / hidden
- For masked versions, maybe they should simply be 'transparent' for standard users (i.e. they don't even see there is a masked version). Or maybe it's better to show that a version has been masked. Not certain about what's the best here

src/components/generics/links/MaskLink.vue
brunobesson: I'll try and have a look, but a bit later, sorry.

brunobesson:
> For masked versions, maybe they should simply be 'transparent' for standard users (i.e. they don't even see there is a masked version). Or maybe it's better to show that a version has been masked. Not certain about what's the best here

Definitely option 1.

src/components/generics/links/MaskLink.vue
asaunier: Thanks! No hurry :)

asaunier:
@lbesson @brunobesson I have more or less followed the specs available in the following document:
[CDC Masquage des versions.docx.pdf](https://github.com/c2corg/c2c_ui/files/9933345/CDC.Masquage.des.versions.docx.pdf)
The latter asks that masked versions are actually listed in the history page but can only be displayed to moderators.

I have made a few simplifications compared to the document though:

1. it's not possible to mask the current version (latest version) => one has to edit it to create a new version, and than it's possible to mask it.
2. I have skipped the version masking logs part (both in the UI and API PRs). It's doable but I don't want to spend too much time on this feature and prefer to stick to a simpler version of the tool.
3. Versions are masked/unmasked one by one, not as a batch => no checkboxes to select the versions to be masked, only a Mask/Unmask link for each version.

As for now what the current UI PR does:

1. mark version as masked (for any user) + add "Mask"/"Unmask" links (for moderators only)
2. masked version links are still available and clickable. Except for moderator they only show an empty content (the API returns a null document) with a message in the history/diff header saying that the version is masked. 
3. diffing based on a masked version returns nothing relevent since at least one version is empty
4. it would make sense to remove the diff links for masked version (except for moderators) but in the UI they are built using the current version id and a "prev" keyword => the "masked" status of the "prev" version is not easily available I think

I admit it the PR could be improved but I don't want to make something too complicated (especially since I am not very familiar with the UI) :)

src/components/generics/links/MaskLink.vue
brunobesson: I do think (not totally certain though) this is a mix of 2 problems:

1. A component isn't supposed to modify the props, it's the responsibility of it's parent, which owns them, usually as data.
2. Vue2 reactivity system is limited to some edge cases, e.g. it cannot detect some array changes (because version is an item of an array).

Anyway, I do think this is more correct:

1. MaskLink component calls the API
2. MaskLink emits an event on success
3. HisoryView listens to event and updates data accordingly (it will then be propagated to MaskLink through props)

If the reactivity still doesn't work, have a look at https://v2.vuejs.org/v2/guide/reactivity.html#For-Arrays

brunobesson: (I could propose code changes and test if you like, only I fear the need to run the backend 🙂 )

asaunier: Thanks! I'll test this on my side and let you know.

asaunier: I guess I should replace 
```
this.version.masked = !this.version.masked;
```
in the ``toggleMask`` method by
```
this.$emit('version-masking', {
    'masked': !this.version.masked,
    'version': this.version
});
```
but then I have not figured out how to listen to the event in the view!?

On the other that looks a bit awkward and complicated to raise an event to the view whereas it would be more convenient and logical to deal with those changes only in the component (that looks like some internal business of the component).

Actually everything is working (eg. the "This version has been masked" shows up - or hides down - as expected when clicking the toggle link). The only things that are not updated are the "Mask" / "Unmask" labels themselves.

As it's only a moderator tool, I have then simplified the code using "Toggle masking" instead, and that's it: e2c8f0c1be2f5a274d8e889a8a83573259d3603c

brunobesson: Hi @asaunier.
I'll have a look at it and test it on the demo as soon as I can catch @lbesson and have the demo bakcend updated (I have release v6.13.0, but need him to migrae the database).

I took the liberty to rebase and force-push your branch on master, along with a commit to fix linting. Hope this doesn't bother you.

brunobesson: It appears that `v-translate` directive with `v-if` is the cause of the problem. It can easily be circumvented ,I didn't dig the root cause.

My remark on not modifying a prop is still valid (the reactivity system proved innocent), but I understand your point.

brunobesson:
@asaunier I took the liberty to work on the subject, if you do not agree, feel free to tash my commits.

What I did:
* hide masked versions in history view for non moderators.
* for moderators, move (un)mask action to its own column, use a iconed button, stripe masked rows for moderators
* it's still possible for a non-moderator to request a masked version, or a diff between 2 versions, of which ate least one is masked. I proposed 2 dialogs in such case

See screenshots below. Wha do you think?

image
image
image

Note that the demo was updated to v6.13.0, which allows to easily test the feature.

asaunier:
@brunobesson I have not tested your changes on my local instance yet but according to your description and screenshots that looks great! Thank you so much :) Feel free to remove all my changes if not needed!

brunobesson:
@asaunier you can test it there: https://c2corg.github.io/c2c_ui/mask_version/# (e.g. https://c2corg.github.io/c2c_ui/mask_version/#/routes/history/53874/fr).

Be sure to play with the demo, not the prod API
image

Issue #2552, opened 2022-02-17
Feature: Boulder grade with uppercase


AntoineMarnat:
Boulder grades should have letters with uppercase, to emphasize the difference with sportclimbing.

See
https://fr.wikipedia.org/wiki/Cotation_en_escalade#Bloc

florentcadot:
Issue can be closed if fine (seems good in production, example: https://www.camptocamp.org/routes/1525585/fr/l-elephant-sein-de-glace letter '7A' grade with letter capitalized).

PR #2971, opened 2022-10-13
Capitalize bouldering cotations

lechinoix:
First contribution to answer to this issue : https://github.com/c2corg/c2c_ui/issues/2552

Made me discover that Fontainebleau boulders are quite well contributed, thanks guys !!

Fontainebleau boulders 😬 
image

Ailefroide multi-pitches ❤️ 
image

### DIFF
--- a/src/components/generics/DocumentRating.vue
+++ b/src/components/generics/DocumentRating.vue
@@ -11,7 +11,7 @@
 
       <span>
         <span v-if="document.rock_free_rating" :title="$gettext('rock_free_rating')">
-          {{ document.rock_free_rating
+          {{ formatBoulderRating(document.rock_free_rating)
           }}<!--
               --></span
         ><!--
@@ -163,6 +163,11 @@ export default {
         },
       };
     },
+	formatBoulderRating(rating) {
+		if (this.document.climbing_outdoor_type !== 'bloc') return rating;
+
+		return rating.toUpperCase();
+	},
   },
 };
 </script>


src/components/generics/DocumentRating.vue
```diff
@@ -163,6 +163,11 @@ export default {
         },
       };
     },
+	formatBoulderRating(rating) {
+		if (this.document.climbing_outdoor_type !== 'bloc') return rating;
```
lechinoix, line 167: I would say that it is not so good to have this hardcoded `'bloc'` here, would be better to have a centralized variable imported in components, but it would be quite a refactoring I believe ^^'

brunobesson:
Thanks for the PR!
Indeed, there is a technical choice to be made beween updating the API and circumventing the 'bug' by capitalizing gradings for bouldering.
I would like to have advice from @cbeauchesne @asaunier or @lbesson to decide which solution is best in their mind.

Anyway, your PR does not handle all cases, e.g. when editing. It could be done quite easily on the UI side, too.
As for the technical part, the grade should rather be a computed property than a method that is called for each rendering.

Note also that we us prettier for formatting the code, you should take this into account when committing.

lechinoix:
Thanks for the review @brunobesson! Yeah, adding a new cotation field for bouldering in the API sounds right, but it would be way more complex than just doing this.

I do not know the codebase so well yet, but probably it means to update contribution form, create a section in the "cotation explained page" and link it in the cotation popup, migrate existing data, add a new filter in search page etc... We could even go as far as create a separate activity actually !

I just wonder if it is really relevant to spend so much effort on bouldering (and who decide what effort we do on it)

lechinoix:
And for the edition part, I just took a look at the code and it seems a little bit too sketchy without help : if we want to keep it on UI side only, we probably need to add a notion of "format" function on UI side in the FormInput that does not modify the value sent to the back for saving (or we would have to be case insensitive backend for ratings).

Sounds like it would be much more clear with a bouldering_cotation saved in backend, what do you think ?

The direct impact of making this change would be that when searching a climbing route by grade, we would have 2 different filters for boulder and lead

lechinoix:
Okay so my tech strat would be something like :

- Adding bouldering_ratings in [common.json](https://github.com/c2corg/c2c_ui/blob/master/src/js/constants/common.json#L68)
- Adding a `bouldering_rating` in [fieldsProperties.json](https://github.com/c2corg/c2c_ui/blob/master/src/js/constants/fieldsProperties.json#L737)
- Setting it visible only for bouldering and hide others in [Field.js](https://github.com/c2corg/c2c_ui/blob/master/src/js/constants/Field.js#L65)
- Adding the new field in RouteEdition and OutingEdition

But I would first need to create bouldering_rating column

brunobesson:
Sorry if my question mislead you @lechinoix.
I was asking for feedback from other developers, but I personaly am convinced we should stick to the UI modification part. Let's wait from other feedbacks if there are, and then decide how to proceed.

asaunier:
I have not thought much at all the implications but as I said in https://github.com/c2corg/v6_api/pull/1420#issuecomment-1288089847 I wonder if this would not be enough to keep using the standard rock climbing ratings and only capitalize them in case of bouldering "routes".

BorutK:
Hi! 
I allready posted this comment in issue #1420. Since I'm not sure if that was the right place or not to make the comment, here goes again:

The point (and probably a problem) is not so much about capitalizing or not. The specificity of Bleau bouldering grades (which ARE accepted since they even were the start for any kind of purely climbing grades) is the fact that the "letter following a number system" only starts with 6a (sorry 6A). 
So from 1 to 5 it's : **1-** , **1** , **1+** , **2-** , **2** , etc.
I on purpose gave the example with 1 because wikipedia in its comparison table puts grades 1, 2 and 3 in a bundle, which is not what is found in real life. Those low grades (with only 3 subdivisions) are important to point out : think of small kids or absolute greenhorns.
In any case, it's nice to see that people care about all this.

Thx to all a ya !

Borut

brunobesson:
Closing, as fixed by #3700

Issue #2939, opened 2022-09-27
Bug: performance regression on YETI


iamvdo:
**The bug**
I noticed a performance regression on YETI (maybe elsewhere too), since the update of openlayers.
After research, it seems that the bug is here since ol@6.13.

The point is, maybe openlayers is not responsible for that, so I investigated further and found out the problem is with ol@6.13, but only using Vue@2 and adding interactions to the map. 
It means that these are working fine configurations:
- ol@6.12 + vue@2
- ol@6.13 + vue@3

Going deeper with openlayers code, I found out that removing [this line](https://github.com/openlayers/openlayers/blob/5e8223ddcb0b55c92345f86cd4418a0ec474fb0f/src/ol/renderer/canvas/Layer.js#L265) fix the problem...

So, my question is quite simple. How can I fix this issue?

- find a workaround (but for now, I'm stuck...)
- revert to ol@6.12 for c2c? (but not reliable on the long run)
- update to vue@3 (but need a lot of work)
- open an issue on openlayers? (why they add this line? mentioning the perf problem with vue@2 and in our specific case)

**To Reproduce**

1. Go to `/yeti/1190971` page (it will load this route in the map)
2. Click `Course` tab
3. Hover over/out the line chunk (3213 points)
4. See route's style update on map

**Expected behavior**
When hover/out the line chunk, the route's style update is nearly instant with ol@6.12, but takes ~1/2 seconds with ol@6.13. With many line chunks, it becomes really unusable.

PS: regarding the workaround solution, I can elaborate further on solutions I already tested, and perf improvements already implemented.

brunobesson:
Hello @iamvdo 
While I sometimes use yeti, I am only using basic functionalities, especially not the route part. So excuse me if I say stupid things.
I cannot reproduce the bug you describe in production: the is nothing happening on the route line on hover, whatever the time I stay on it.

That said:
* indeed, migrating to vue3 is a task we have to forecast, but it won't happen soon
* sticking to ol@6, we could, as long as there is no security issue occurring. And indeed, it's not a viable solution in the future

So what I propose, is that you first raise an issue by ol team first, see how they respond. Depending on their answer, we will have more clues about what our best path is.

iamvdo:
Thanks @brunobesson!
I opened a bug on openlayers repo, let's see how it goes :)
https://github.com/openlayers/openlayers/issues/14151

Also, I made videos to illustrate the bug

[...]

PR #3008, opened 2022-10-24
[YETI] Fix performance issue

iamvdo:
This fixes issue #2939.

It moves cartoLayers and dataLayers outside of Vue's reactivity system.


src/components/yeti/YetiMap.vue
```diff
@@ -184,6 +177,16 @@ export default {
     showAreas() {
       this.updateCartoLayersOpacity();
     },
+    visibleCartoLayerId(id) {
+      console.log('visibleCartoLayerId')
```
brunobesson, line 181: please no console statements 🙂

brunobesson (approved):

src/components/yeti/YetiMap.vue
iamvdo: :facepalm: 
Done!

PR #3009, opened 2022-10-24
feat(statistics): fix NaN when there is no outing yet

mathieudutour:
Saw this 
Screenshot 2022-10-24 at 14 48 31
and decided to fix it directly

### DIFF
--- a/src/views/portals/outings-stats/OutingsStatsView.vue
+++ b/src/views/portals/outings-stats/OutingsStatsView.vue
@@ -73,7 +73,11 @@ export default {
     },
 
     progress(current, total) {
-      this.loadingPercentage = current / Math.min(total, LIST_MAX_LENGTH);
+      if (!total) {
+        this.loadingPercentage = 0;
+      } else {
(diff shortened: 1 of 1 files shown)


brunobesson:
Thanks for your PR @mathieudutour !

If a user has no outings, the case is not handled very well yet, I agree.
However, I am not sure 0% is he correct choice. Should'n it be 100% instead?
Also, we could avoid displaying empty graph and print a small text instead (but this could be another work).

What do you think?

brunobesson:
> True

Would you volunteer to continue on this issue (i.e. hide empty graphs)?

mathieudutour:
Happy to yes, will do so in another PR

mathieudutour:
Do we have an existing screen with an empty state (so that I don't reinvent the wheel)?

brunobesson:
> Do we have an existing screen with an empty state (so that I don't reinvent the wheel)?

I cannot think on any case for this, except maye empy results in a table:
image

Anyway, I do not think of something very complicated: a simple paragraph with some style saying e.g. `No outings yet. Share your achievements and feedback with the camptocamp community to have some stats!`

PR #3016, opened 2022-10-26
Support webp and avif format via <picture>

lbesson:
The goal is to replace \ by \<picture> each time we refer to a thumbnail (SI, MI, BI) of a c2c image, and propose webp and avif format to the browser.

A timestamp and start id are defined in config, so that we don't try to get avif or webp if they are not generated.
Ultimately we should have a script generating all missing avif and webp thumbnails and get rid of this limitation.

There are a few places to be checked (currently slideshows are often broken):

- [x] AssociationImageLink
- [ ] Gallery - often a display / slide issue until we resize a bit the window with chrome ??
- [x] Markdown
- [x] ImageViewer
- [x] ImageCards
- [x] SophiePictureContestView
- [x] DiffView
- [x] ImageEditionView

(All of the above only tested on chrome desktop and firefox desktop at the moment)


brunobesson:
@lbesson what is the status on this PR? do you need help?

brunobesson:
Replaced by #3111

PR #3022, opened 2022-10-27
Correctly display f-number and exposure time

lbesson:
Currently, an error is thrown on image view when trying to display the f-number
(see for example https://www.camptocamp.org/images/1438128/fr/ and look at the console)

This PR fixes the error by expliciting that the fnumber is a number (!) and should not be translated. It also adds a "prefix" option for field properties in order to display "f/2" for example.

PS: the translation for fnumber in french is "Propriétés de l'image", which is incorrect (Should be "Ouverture"). I reported an issue in transifex, but I'm not used to the tool, and didn't check the other languages.

### DIFF
--- a/src/components/generics/DocumentField.vue
+++ b/src/components/generics/DocumentField.vue
@@ -27,7 +27,10 @@
  --><span v-else-if="field.i18n">{{ $gettext(value, field.i18nContext) }} </span
     ><!--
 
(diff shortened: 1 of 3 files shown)


lbesson:
fixes #3023 and #3024

lbesson:
image
Seems translation in other languages aren't perfect either, at least in english

Issue #2451, opened 2022-01-06
Feature: add multi-criteria search filter on outing list (mobile)


loicperrin:
**Community discussion about this feature (mandatory)**
https://forum.camptocamp.org/t/mobile-filtres-sur-la-liste-de-sorties/278156

**Is your feature request related to a problem? Please describe.**
On mobile, add multi-criteria search filter on outing list. This field already exists on desktop UI and is very usefull to filter by Massif

PR #3051, opened 2022-11-08
feat: add multi-search dropdown in mobile

jeanphilippeds:
Hello,

A small PR to fix https://github.com/c2corg/c2c_ui/issues/2451

With selected tags:

With multi-search results:

I'm not a big fan of repeating the `filter` icon. Maybe we could put another one.

### DIFF
--- a/src/views/documents/utils/QueryItems.vue
+++ b/src/views/documents/utils/QueryItems.vue
@@ -38,6 +38,23 @@
           />
         </div>
       </dropdown-button>
+      <dropdown-button key="multi-search" class="query-item-component is-hidden-tablet">
+        <span slot="button" class="button is-size-7-mobile">
+          <fa-icon :icon="$options.categoryIcon['MultiSearch']" />
+          <span>&nbsp;</span>
+          <fa-icon icon="angle-down" aria-hidden="true" />
+        </span>
+        <div class="sub-query-items">
+          <association-query-item
+            class="dropdown-item"
+            :document-types="associations"
+            @add="addTag"
+          />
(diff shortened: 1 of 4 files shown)


src/views/documents/utils/QueryItems.vue
```diff
@@ -17,17 +17,12 @@
         <span slot="button" class="button is-size-7-mobile" :disabled="category.fields.length === 0">
           <fa-icon :icon="$options.categoryIcon[category.name]" />
           <span class="is-hidden-mobile">
-            <!-- $gettext('General') -->
-            <!-- $gettext('ratings') -->
-            <!-- $gettext('Terrain') -->
-            <!-- $gettext('Miscs') -->
```
brunobesson, line 23: Is thee a specific reason for removing these comments? They are i18n-related, and should not be removed (except if unused). See https://github.com/c2corg/c2c_ui/wiki/i18n

brunobesson (approved):
Other than the removed i18n stuff, looks good to me.

src/views/documents/utils/QueryItems.vue
jeanphilippeds: Ok my bad, I thought it was dead code

brunobesson:
There might be some problems when the criteria title is larger than the phone width:

image

jeanphilippeds:
Good catch!
I've added some commits. I also fixed the margins between tags when we have multiple lines.

brunobesson:
Nice job! I'll merge.

Issue #3045, opened 2022-11-08
Feature: Improve naming of gpx/kml traces downloaded from camptocamp


anakreenskyrider:
**Community discussion about this feature (mandatory)**
[Link to <https://forum.camptocamp.org/> where this feature has been validated](https://forum.camptocamp.org/t/nom-des-traces-gpx/307603/11) https://forum.camptocamp.org/t/nom-des-traces-gpx/307603/12

**Is your feature request related to a problem? Please describe.**
When downloading the gpx (or kml) traces from outing s or routes, it's named after the document ID only, which makes it difficult to discriminates within multiples traces dowloaded. 

**Describe the solution you'd like**
For routes : automatically name the gpx (or kml) file with document ID + name of the summit + the name of the route
For outings : automatically name the gpx (or kml) file with document ID + name of the outing + date realized

**Describe alternatives you've considered**
maybe remove doc ID if name of the file is too long ?

**Additional context**
N/A

lbesson:
https://github.com/c2corg/c2c_ui/blob/1bc9d7d53b4210245af84ec3d466d0642a57939f/src/views/document/utils/boxes/MapBox.vue#L145

anakreenskyrider:
thanks !!!

PR #3074, opened 2022-11-17
feat: improve naming of downloaded traces 

TitouanLarnicol:
Issue #3045 Renaming of the downloaded files: 

- For routes :  document ID + name of the summit + the name of the route
- For outing : document ID + name of the outing + date realized

I implemented the alternative described by [anakreenskyrider](https://github.com/anakreenskyrider) which remove the ID if the name if longer than 100 characters ( decided completely arbitrarily, might be too long ? )

### DIFF
--- a/src/views/document/utils/boxes/MapBox.vue
+++ b/src/views/document/utils/boxes/MapBox.vue
@@ -142,12 +142,21 @@ export default {
 
       feature.set('name', name);
 
-      const filename = this.document.document_id + extension;
+      const filename = this.getDownloadedFileName(name, extension);
       const content = format.writeFeatures([feature], {
         featureProjection: 'EPSG:3857',
       });
 
       utils.download(content, filename, mimetype + ';charset=utf-8');
+    },
+
+    getDownloadedFileName(name, extension) {
+      let filename = this.document.document_id + '_' + name.substring(0, 100);
+      if (this.documentType === 'outing') {
+        filename = filename + '_' + this.document.date_start;
+      }
+      // Remove forbidden characters/spaces, and replace them with '_'
+      return filename.replace(/[/\\?%*:|"<>]/g, ' ').replace(/\s+/g, '_') + extension;
     },
   },
 };


brunobesson (commented):
All good for me, thanks @TitouanLarnicol!
I wonder if maybe we should handle spaces and special characters and replace them with e.g. underscores?
I assume it is platform and browser dependent, but I get e.g. `1483583_Grande Moucherolle _ Traversée des Moucherolles_2022-11-13.gpx`.
I feel something like `1483583_Grande_Moucherolle_Traversée_des_Moucherolles_2022-11-13.gpx` would be more appropriate? What do you think?

TitouanLarnicol:
Good idea @brunobesson , i will add that to the PR.

src/views/document/utils/boxes/MapBox.vue
```diff
@@ -142,13 +142,22 @@ export default {
...
+      const filename = this.getDownloadedFileName(name, extension);
       const content = format.writeFeatures([feature], {
         featureProjection: 'EPSG:3857',
       });
 
       utils.download(content, filename, mimetype + ';charset=utf-8');
     },
+
+    getDownloadedFileName(name, extension) {
+      const prefixRouteId = name.length < 100 ? this.document.document_id + '_' : '';
```
asaunier, line 154: That would be more consistent to always prefix with the route id but substring the whole name at the first 100 characters if really to long?

asaunier: `prefixDocumentId` would be more relevant as the document may be an outing :P

TitouanLarnicol: > That would be more consistent to always prefix with the route id but substring the whole name at the first 100 characters if really to long?

it would be better indeed

TitouanLarnicol:
I let you check the changes. 
Added a little regex because there is always a ":" in the name of a route. I converted illegal arguments into spaces and replace every space with '_' to avoid names like : "Mont_Rachais___face_Est" (Browsers convert them into '_').
I can also remove the regex if you think it's better to let chrome adding those '_'.

brunobesson (approved):
Looks good to me, thanks!

PR #3111, opened 2022-12-07
Support webp and avif format via <picture>

brunobesson:
The goal is to replace \ by \<picture> each time we refer to a thumbnail (SI, MI, BI) of a c2c image, and propose webp and avif format to the browser.

A timestamp and start id are defined in config, so that we don't try to get avif or webp if they are not generated.
Ultimately we should have a script generating all missing avif and webp thumbnails and get rid of this limitation.

There are a few places to be checked (currently slideshows are often broken):

- [x] AssociationImageLink
- [x] Gallery - often a display / slide issue until we resize a bit the window with chrome ??
- [x] Markdown
- [x] ImageViewer : error when zooming in / out
- [x] ImageCards
- [x] SophiePictureContestView
- [x] DiffView
- [x] ImageEditionView

(All of the above only tested on chrome desktop and firefox desktop at the moment)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "c2c_ui",
       "version": "7.14.0",
+      "hasInstallScript": true,
       "license": "AGPL-3.0-or-later",
       "dependencies": {
         "@brunobesson/vue-awesome-swiper": "4.1.4",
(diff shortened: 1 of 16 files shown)


brunobesson:
Last found bugs should be fixed now by patching swiper 😒

@lbesson could you review and test?

lbesson:
@brunobesson I changed the timeout from 100 to 0 and provided additional comment.
Can you please (again) test on your side and if happy, merge?

lbesson (approved):

brunobesson:
If the problem resides only in the gallery spacing between images, then yes it seems to b fine.

lbesson:
hum, I should have squashed

PR #3114, opened 2022-12-08
Yeti updates

iamvdo:
Hello,

Small updates on the YETI app.
It fixes a bug on map layers (blinking) when navigating to/from other pages.
It also prepares the addition of Swiss avalanche bulletin data.

Thanks for review!
Vincent


src/components/yeti/map-layers/AvalancheBulletinsLayer.vue
```diff
@@ -308,7 +308,8 @@ export default {
       let mountainsFeatures = this.getFeaturesFromData(data);
       // TMP: filter only France for now
```
brunobesson, line 309: this comment is now inconsistent with the logic below

brunobesson (approved):
Not much to say 🙂

Issue #3091, opened 2022-11-29
Feature: Images should be above hline coming from sections


AntoineMarnat:
**Community discussion about this feature (mandatory)**
Link to <https://forum.camptocamp.org/t/separations-pas-tres-heureuses/308456>, where this feature has been mentioned.

**Is your feature request related to a problem? Please describe.**
See forum : hline are above the images. It should be the opposite.

AntoineMarnat:
The fix does not work as expected and brakes the presentation of a lot of documents.

PR #3115, opened 2022-12-08
fix(markdown): add style to markdown content for better images placement

NinjaKinshasa:
fix https://github.com/c2corg/c2c_ui/issues/3091

* style of titles changed from block to inline-block + width 100% to avoid overlapping with images
* style of markdown-content changed to overflow : auto to prevent images to overflow the markdown container

### DIFF
--- a/src/components/generics/Markdown.vue
+++ b/src/components/generics/Markdown.vue
@@ -166,10 +166,17 @@ export default {
 <style lang="scss">
 // Not scoped syle, because CSS selector are not explicitly present in template
 .markdown-content {
+
+  overflow: auto;
+
   &:not(:last-child) {
     margin-bottom: 1.5rem;
   }
 
+  h3, h4, h5 {
+    display: inline-block;
+    width: 100%;
+  }
   h3 {
     font-size: 1.5rem !important;
     margin-bottom: 0.5em !important;


brunobesson:
Thanks for your PR!

I see a problem : images are positioned to the right, but "out of content", while the content should give space to the image.
Similar to word-like documents image positioning.

NinjaKinshasa:
Thank you for your reply, can you provide a visual exemple of the expected result, or a screen of the problem with the result generated with my code ?
If I understand your comment, your concern is that the image is displayed under the text instead of being displayed on the same row. Actually, the image goes under the text only when the image tag is under the text in the original markdown. To display the image and the text on the same row, the image tag should be above the text in the markdown, like in html/css.

Here is an exemple : 

title (1)
img (2)
text (3)
img (4)
text (5)
title (6)
img (7)
text (8)

prev

I  understand that merging this PR will probably mess up the appearance of some publications, but I do not see a better tradeoff to fix the current issue while only applying css to markdown code. Do you think to something else ?

brunobesson:
Indeed, my bad!
I assumed another approach, so didn't think of the image positioning in markdown code.

Everything's fine for me then.

PR #3123, opened 2022-12-13
Allow to open map feature in new tab/window

brunobesson:
Using `CTRL` key, feature is opened in new tab/window (depending on browser configuration). User may need to accept popups opening first.

In the same time, do not open document if self (be it in another tab/window on in place)


lbesson:
that's when clicking on a map feature (e.g. summit)?

brunobesson:
> that's when clicking on a map feature (e.g. summit)?

indeed

lbesson (approved):

PR #3136, opened 2022-12-19
chore: fix prettier configuration for .vue files

### LINKED ISSUE #3135: Question: new linting rules?
Hello,
I noticed that the linting is not as strict as before (semicolon/commas can be omitted, new line at the end of line not mandatory, indentation not respected is ok, etc.)

Is it intended? Or just default behaviour that changed? Or a problem in my side?

Thanks!

brunobesson:
Fixes #3135

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,12 @@
   "env": {
     "node": true
   },
-  "extends": ["plugin:vue/recommended", "plugin:import/errors", "plugin:import/warnings", "prettier"],
+  "extends": [
+    "plugin:vue/recommended",
+    "plugin:import/errors",
+    "plugin:import/warnings",
+    "plugin:prettier/recommended"
+  ],
   "reportUnusedDisableDirectives": true,
   "rules": {
     "no-unused-vars": "warn",
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,5 @@
 src/assets/font-awesome-custom/
 src/translations/
+dist/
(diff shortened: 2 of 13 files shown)


iamvdo (approved):
lgtm :)

iamvdo:
It seems that lint does not occur on save (while `npm run serve`)?

brunobesson:
> It seems that lint does not occur on save (while npm run serve)?

Well, unless I am saying something stupid, prettier on save depends on your IDE and plugin configuration. Programmatically, it is only configured to lint with CI (`npm run lint:no-fix`), or when commiting files (with `husky` and `pretty-quick`). Seems OK for me with e.g. vscode?

iamvdo:
OK, a bit different workflow than before, but I configure it with no problem.
Thanks!

brunobesson:
Just to know: you mean that before, `vue-cli` applied linting on run? I don't remember / hink it ever did?

iamvdo:
Yes, all lint errors were in the console. Right now, only the eslint ones are, not prettier.

brunobesson:
OK then, thanks for the feedback.
I haven't time to dig deep into it, but I guess it broke when I update `vue-cli` to v5. I made a commit to enable `eslint-plugin-prettier` which should run `prettier`along with `eslint`.
Does it provides you back with the `prettier` errors when serving @iamvdo ?

iamvdo:
Yes, it works as before! Thanks

Issue #3132, opened 2022-12-19
Feature: add catalan maps


brunobesson:
http://www.icc.cat/vissir3/

https://www.icgc.cat/en/Web/Help/Legal-notice
https://www.icgc.cat/en/Public-Administration-and-Enterprises/Services/Online-services-Geoservices/WMS-and-tiles-Reference-cartography/WMS-Current-maps-and-orthophotos

Problem: rework layer selector UI, there is no more place!

PR #3142, opened 2022-12-20
feat: rework layer selector

brunobesson:
As more sources get included, we need to rework it.

image
image
image


brunobesson:
Fixes #3132

PR #3188, opened 2023-01-05
feat: add Slovenian as API language

brunobesson:


lbesson (commented):
I guess if you looked for all occurences that's fine?

PR #3208, opened 2023-01-13
YETI updates

iamvdo:
This PR fixes some bugs and adds:

- new style for area layer (yeti extent), yeti layer (server image) and slopes. It uses compositing and blend modes.
- new styles for avalanche bulletin (fill area, strokes, etc)
- elevation profile box for route

Thanks for review!
Vincent


iamvdo:
Hello,
Sorry for force-push. I forgot to add `package-lock.json`, but after updating modules in my side, there was so many changes, so I had to rebase from master.

src/components/yeti/map-layers/simplify.js
```diff
@@ -0,0 +1,124 @@
+/*
```
brunobesson, line 1: why is i copied, and not importing the npm pakage instead?

src/components/yeti/ElevationProfile.vue
```diff
@@ -0,0 +1,386 @@
...
+      Yetix.$emit('elevationProfile', 'cursor_move', coord);
+    },
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/sass/variables';
+
+$C2C-orange: red;
```
brunobesson, line 323: not sure to understand this one?

src/components/yeti/ElevationProfile.vue
```diff
@@ -0,0 +1,386 @@
+<template>
```
brunobesson, line 1: there are diffs between c2c's elevation profile and this one, but maybe we could build some factorized code? I assume you thought of it bu voted against?

brunobesson (commented):
I confess I didn't go deep into compositing and blend modes, just checked the rendering. Would appreciate 2 lines to explain the choices and impact though 👍 Shortly, the mode you zoom, the lighter the layer on top?
I especially like the new rendering for avanche bulletins.

As for the elevation profile, it is not computed when drawing, thus only applicable when importing a file?

Two questions came to my mind by checking your nice work:

* maybe we could try to unify the layer chooser design? I rewrote it on c2c side because we couldn't handle so mny layers on mobile mode. This is not required on your side (yet? 🙂), but still?
* I am wondering for a long time if we could include some of your work within c2c carto. I am mainly thinking of avalanche bulletins, nivose, Romma, FlowCapt. Are there right restrictions?

iamvdo:
Hi Bruno.

Thanks for review.

Regarding **compositing and blend modes**, it offers better colors when multiple layers are stacked. For example, slopes layers pixels are multiplied with pixel from layers below them, and so opacity can be increased. Same for Yeti layer (computation image). 
For area-layer (Yeti extent), pixels are drawn only through the shape (so, it seems clipped) and the effect slowly fade away as the zoom increases.
For avalanche bulletins, no compositing, it is just drawn on top of that. For their styles, i construct inner strokes based on a buffered geometry (we're currently trying to simplify some shapes from the server side)
Performance wise, we can notice a small impact when layers are all active, but i think it is acceptable (let me know if you feel the UI is slow).

For **elevation profile**, yes for now, only imported file or route from camptocamp have elevation. We've build an API to get 3D points when drawing, and the next part will be to implement it from the client.

For the **layer chooser**, yes I noticed the change on c2c, and I need to work on it for Yeti too. Maybe it's a great idea to unify those yes.

For **including some yeti layers inside c2c**, I personally have no opinion. Maybe not a straightforward task as OLMap and YetiMap diverged a while ago. Maybe @jbeilin can answer with another point of view?

Vincent

package.json
```diff
@@ -27,6 +27,8 @@
     "@fortawesome/free-regular-svg-icons": "6.2.1",
     "@fortawesome/free-solid-svg-icons": "6.2.1",
     "@fortawesome/vue-fontawesome": "2.0.9",
+    "@turf/buffer": "^6.5.0",
+    "@turf/difference": "^6.5.0",
```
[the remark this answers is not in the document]
iamvdo: Done

src/components/yeti/ElevationProfile.vue
iamvdo: Me neither :P
I admit that for elevation, I started from the c2c Elevation's component, and build what I wanted without too much attention to details.
(next conversation) => Yes, I assumed that this would be 2 different things (yeti elevation doesn't have duration for example), and so, build our own component, as for the map. But I understand it can be problematic in the long time (we see that for the layer chooser on the map)
So, what's your point?

jbeilin:
> Hi Bruno.
> 
> Thanks for review.
> 
> Regarding **compositing and blend modes**, it offers better colors when multiple layers are stacked. For example, slopes layers pixels are multiplied with pixel from layers below them, and so opacity can be increased. Same for Yeti layer (computation image). For area-layer (Yeti extent), pixels are drawn only through the shape (so, it seems clipped) and the effect slowly fade away as the zoom increases. For avalanche bulletins, no compositing, it is just drawn on top of that. For their styles, i construct inner strokes based on a buffered geometry (we're currently trying to simplify some shapes from the server side) Performance wise, we can notice a small impact when layers are all active, but i think it is acceptable (let me know if you feel the UI is slow).
> 
> For **elevation profile**, yes for now, only imported file or route from camptocamp have elevation. We've build an API to get 3D points when drawing, and the next part will be to implement it from the client.
> 
> For the **layer chooser**, yes I noticed the change on c2c, and I need to work on it for Yeti too. Maybe it's a great idea to unify those yes.
> 
> For **including some yeti layers inside c2c**, I personally have no opinion. Maybe not a straightforward task as OLMap and YetiMap diverged a while ago. Maybe @jbeilin can answer with another point of view?
> 
> Vincent

Hi Bruno and Vincent, 
In my opinion, Yeti can be seen as the "avalanche page" in CampToCamp. Therefore, it might be better to keep the elements specific to this problem in this page. Avalanches bulletins, nivoses, flowcapt are mainly useful for winter activities, not for all outdoor activities.  
Jacques

brunobesson:
I don' ave time to answer to all your feedback right now, so just a point for now:

> Avalanches bulletins, nivoses, flowcapt are mainly useful for winter activities, not for all outdoor activities.

The same could somehow be said for slopes, isn't it?

src/components/yeti/map-layers/simplify.js
iamvdo: Well, it seems that package is 2d only, and it uses object values (x/y) instead of list. The script is not configurable to maintain performance. The advice is to copy/paste and update as needed.

See http://mourner.github.io/simplify-js/
> Configurability of point format would draw a significant performance overhead, so if you use a different format, e.g. [Number, Number], just run search/replace of .x and .y through the source of the library to suite it to your application. For a version that works with 3D points, check out the 3d branch.

brunobesson: OK, didn't notice it. Fine then!

src/components/yeti/ElevationProfile.vue
brunobesson: Indeed, it's the same in @lbesson's code. I guess it was temporay and not cleaned tereafter?
@iamvdo @lbesson what would you think of using the c2c orange ($color-base-c2c) for the elevation profile?

brunobesson: image

brunobesson: As for factorization, it's not a priority, but it could be enanced someday to reduce our bundle sizes.

jbeilin:
> I don' ave time to answer to all your feedback right now, so just a point for now:
> 
> > Avalanches bulletins, nivoses, flowcapt are mainly useful for winter activities, not for all outdoor activities.
> 
> The same could somehow be said for slopes, isn't it?

Yes, sure. I think this should be discussed but the scope is a bit larger then the strict perimeter of this PR.

src/components/yeti/ElevationProfile.vue
iamvdo: I've found the red a good choice here, because it matches the related route's color on map too (yellow and red when hover/selected). That's why I keep it as-is.

iamvdo:
Maybe a silly idea (I'm not practising mountain activities, so not a c2c user), but why not add "Yeti" or "Avalanche Risks" directly in the left menu? (As for "Outings" or "SERAC"). So people can quickly choose the most appropriate map based on their activities.
It does not exclude the fact that layers can be shared between maps.

brunobesson:
> Maybe a silly idea (I'm not practising mountain activities, so not a c2c user), but why not add "Yeti" or "Avalanche Risks" directly in the left menu? (As for "Outings" or "SERAC"). So people can quickly choose the most appropriate map based on their activities.
It does not exclude the fact that layers can be shared between maps.

I do agree with providing more visibility to Yeti. The left menu is a tricky part however, beause of the ad. We do not have much space and there were many discussions about what we can put in the left menu while allowing the ad to be displayed on most devices. This could/should be a question asked to board@c2c and dev@c2c, to relaunch the debate.

PR #3264, opened 2023-02-01
fix: update language option for Chinese

hemind:
change language option 中文(means Chinese) to more accurate 简体中文(means Simplified Chinese)


hemind:
already done by updating from transifex

PR #3338, opened 2023-03-06
Update TrackersView.vue

AntoineMarnat:
Update link to Outdoorvision with tracker

### DIFF
--- a/src/views/user/TrackersView.vue
+++ b/src/views/user/TrackersView.vue
@@ -37,7 +37,7 @@
           preservation and design of my practice sites in France.
         </p>
         <div class="has-text-centered">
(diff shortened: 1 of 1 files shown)


brunobesson:
What's the use of `noreferrer noopener` then? ❓

AntoineMarnat:
> What's the use of `noreferrer noopener` then? ❓

Sorry, I just changed the link without further checks.

Outdoorvision wishes to have this tracker. I am not a specialist, may it imply some security issues ? Or should I just remove the noreferrer noopener parameter without further considerations ?

brunobesson:
Nah, forget my comment. `noopener noreferrer` aims at protecting privacy by hiding some information to the recipient. But then if we add the query param, this is somehow useless (not totally, we hide most informations still).

Issue #3405, opened 2023-04-02
Homepage - Icons display (dashboard dans feed) should be improved


flob38:
![image](https://user-images.githubusercontent.com/2234024/227933249-a1231ba0-0a79-4b66-920e-bd4faa778310.png)
There is a problem IMHO when there are multiple activities on mobile

_Originally posted by @brunobesson in https://github.com/c2corg/c2c_ui/pull/3335#discussion_r1149196768_

PR #3414, opened 2023-04-05
style: Better display of icons on dashboard view

flob38:
Dashboard part of #3405


flob38:
Not that pretty when screen is "wide-mobile" nor when there are more than 4 activities for 1 doc (however, 1rst April appart, I don't think we see more than 4 activities/doc very often).

brunobesson:
image

Still needs work IMHO 🙂 

I really do not have time to work on this this week, sorry. 
May I suggest trying to put small icons below areas and keep activities on the right ?
Also, do not try to fix the separation between both blocks, let the browser handle the repartition.

brunobesson:
> You're on the feed view

My bad! Still, event on wide screens, there is a problem IMHO with more than 2 activities (I agree, this is quite exceptional). On mobile, the current solution doesn't look that nice either with 2 activities IMHO.

image
image

Have there been discussion on the matter on the forum? I couldn't follow.

flob38:
No real discussion on the forum for those little fixes.
On mobile, the problem shows when title is too long, or when there are a lot of activities. That why I decided to have only 2 activity icons on the same line.
Maybe it should be better to have activity icons and outing icons on another line than the title, on mobile ?

flob38:
Discussion within CA : https://forum.camptocamp.org/t/nouvelle-page-daccueil-version-tableau-de-bord/315091
Ok for this design

PR #3432, opened 2023-04-16
style: Better display of icons on feed view

flob38:
Feed part of https://github.com/c2corg/c2c_ui/issues/3405


flob38:
No real discussion on the forum for those little fixes.
I keep outings icons on the right of activity's, seems more "compact" (in order to display more cards without scrolling).

brunobesson (commented):
I do feel lignment sould be preserved instead. From right to left: 
* there is at most 4 'outing' icons on the right => always leave enough place for those
* then take all the place needed (but not more) for activity icons
* and last the areas text, using all remaining place

flob38:
Discussion within CA : https://forum.camptocamp.org/t/nouvelle-page-daccueil-version-fil-dactivite/315095
Ok with this design

PR #3618, opened 2023-09-05
feat: Add canonical url to front pages

lechinoix:
Adding a link to specify canonical url to google bots. It might be interesting as the page is linked from all other pages with only the /routes/id.

This Pull Request is opened following this forum thread : https://forum.camptocamp.org/t/les-topos-camptocamp-ne-sortent-plus-sur-google/320501/59?u=lechinoix


lechinoix:
I add here that I am not a professional of SEO and that there might be a better choice to set this canonical (for example, should we put the locale in the canonical, should we put the title etc...)

src/views/document/utils/document-view-mixin.js
```diff
@@ -90,6 +90,9 @@ export default {
 
       return [...result, ...this.documentOpenGraph()];
     },
+    link: function () {
+      return [{ rel: 'canonical', href: `https://www.camptocamp.org${this.getCurrentPath()}`, id: 'canonical' }];
```
brunobesson, line 94: what's the use of the id?

brunobesson (approved):
Looks good to me, although I am no SEO expert. What's the use of the `id` attribute?

PR #3691, opened 2023-10-30
fix: SwissTopo retina tiles show up 2x too big

eddy-geek:
before / after

image

Only topo is set as retina, for the others it's wasteful. The screenshot shows that we can still stack retina topo and non-retina slope layers.

adapted from https://gis.stackexchange.com/a/395480/176462


eddy-geek:
superseded by #3695

PR #3700, opened 2023-11-08
feat: #2552 capitalize rock free rating for bloc

florentcadot:
Hi! Here is a PR trying to solve https://github.com/c2corg/c2c_ui/issues/2552

I've taken into account what had been suggested in this PR : https://github.com/c2corg/c2c_ui/pull/2971.  So, here is a solution that is :
- front-end only
- use a computed property to capitalize the boulder grade (in DocumentRatingComponent.vue)
- capitalize the boulder grade field and options in the create/edit form

I wanted to encapsulate the logic to convert sport climbing grade to boulder grade (and boulder grade  -> sport climbing grade) somewhere but I didn't find an appropriate place so I've let it inside the components.   

FYI: I've considered creating a new type of "form field" especially for boulder grade but I think it would have required more changes and a kind of duplication (as boulder grades are just sport climbing grade with a capital letter).

FYI2: I've not been able to persist data locally when I've tried to edit/create a new route (with or without the changes of this PR), I guess some entities schema changed and the docker images had not been updated, but I'm not sure

### DIFF
--- a/src/components/generics/DocumentRating.vue
+++ b/src/components/generics/DocumentRating.vue
@@ -11,7 +11,7 @@
 
       <span>
         <span v-if="document.rock_free_rating" :title="$gettext('rock_free_rating')">
-          {{ document.rock_free_rating
+          {{ rockFreeRating
           }}<!--
               --></span
         ><!--
@@ -148,6 +148,11 @@ export default {
       }
 
       return result;
+    },
+    rockFreeRating() {
+      const rating = this.document.rock_free_rating;
+      if (this.document.climbing_outdoor_type !== 'bloc') return rating;
+      return rating.toUpperCase();
     },
   },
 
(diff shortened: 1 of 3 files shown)


brunobesson (approved):
Looks like an efficient solution to me!

brunobesson:
> FYI2: I've not been able to persist data locally when I've tried to edit/create a new route (with or without the changes of this PR), I guess some entities schema changed and the docker images had not been updated, but I'm not sure

I tested it, it works fine.

florentcadot:
> > FYI2: I've not been able to persist data locally when I've tried to edit/create a new route (with or without the changes of this PR), I guess some entities schema changed and the docker images had not been updated, but I'm not sure
> 
> I tested it, it works fine.

@brunobesson You're completely right, I've forgot to launch the ElasticSearch syncer script !

Issue #3768, opened 2023-12-20
Bug: The Dockerfile uses an outdated OpenResty/Debian image


LaurentBonnaud:
Hi,

the Dockerfile contains this line:

`FROM openresty/openresty:bullseye as production-stage`

Is there a good reason to use `bullseye` (Debian 11) instead of `bookworm` (Debian 12) ?

florentcadot:
Hi @LaurentBonnaud,

This base image had been chosen one year ago (19/12/2022, cf pic below) when bookworm was not released yet (it had been released in June 2023 according to https://en.wikipedia.org/wiki/Debian_version_history).

![image](https://github.com/c2corg/c2c_ui/assets/52289507/25e0c81f-5478-4beb-8c38-5009be288948)

Here is a PR to use bookworm instead: https://github.com/c2corg/c2c_ui/pull/3770 .

LaurentBonnaud:
Thanks for the PR!

florentcadot:
Issue can be closed I guess

PR #3770, opened 2023-12-23
chore: #3768 use openresty debian bookworm as production base docker image

florentcadot:
Hi! Here is a PR to use the last debian version (bookworm) for the openresty base docker image as suggested in this issue https://github.com/c2corg/c2c_ui/issues/3768.

Here is a solution for variable substitution using **sed**  instead of  **envsubst** as it was not provided with this debian version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ RUN node tools/generate-health.js
 
 # production stage
 FROM openresty/openresty:bookworm as production-stage
+RUN apt-get update && apt-get install -y gettext-base # required to use envsubst
 COPY --from=build-stage /dist /usr/share/nginx/html
 COPY ./docker/nginx.conf /etc/nginx/conf.d/default.conf.template
 
@@ -18,8 +19,5 @@ COPY ./docker/nginx.conf /etc/nginx/conf.d/default.conf.template
 ENV PORT 80
 ENV HEALTH_HTTP_PORT 8080
 ENV SERVER_NAME _
-
-# Required to use envsubst
-RUN apt-get update && apt-get install -y gettext-base
 
 CMD ["sh", "-c", "envsubst '${PORT},${HEALTH_HTTP_PORT},${SERVER_NAME}' < /etc/nginx/conf.d/default.conf.template > /etc/nginx/conf.d/default.conf && /usr/bin/openresty -g 'daemon off;'"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN npm run build
 RUN node tools/generate-health.js
 
 # production stage
-FROM openresty/openresty:bullseye as production-stage
+FROM openresty/openresty:bookworm as production-stage
 COPY --from=build-stage /dist /usr/share/nginx/html
 COPY ./docker/nginx.conf /etc/nginx/conf.d/default.conf.template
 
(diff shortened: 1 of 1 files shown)


LaurentBonnaud:
Thanks for this update!
How about installing the `gettext-base` package on top of the image to get the `envsubst` command back?

florentcadot:
> Thanks for this update! How about installing the `gettext-base` package on top of the image to get the `envsubst` command back?

(Thanks for the fast review :smile:) 

You're right, the image is correctly built using `gettext-base` package, just need to add the following line in the Dockerfile :

`RUN apt-get update && apt-get install -y gettext-base
`

But as we can achieve exactly the same result using the `sed` command available without installing anything else it seems "lighter" to me.

I think it is just a question of coding style/approach, no problem for me if there is a consensus about using the `gettext-base` package

LaurentBonnaud:
> But as we can achieve exactly the same result using the sed command available without installing anything else it seems "lighter" to me.

Indeed, lighter, but more difficult to read.

florentcadot:
> > But as we can achieve exactly the same result using the sed command available without installing anything else it seems "lighter" to me.
> 
> Indeed, lighter, but more difficult to read.

Updated installing the `gettext-base` package on top of the image

LaurentBonnaud:
Thanks for this new version!

Dockerfile
```diff
@@ -19,4 +19,7 @@ ENV PORT 80
 ENV HEALTH_HTTP_PORT 8080
 ENV SERVER_NAME _
 
+# Required to use envsubst
+RUN apt-get update && apt-get install -y gettext-base
+
```
brunobesson, line 24: for cache reasons, it is best to run this command as soon as possible, i.e. before line 14.

brunobesson (changes requested):
All good, except some cache optimizations

brunobesson (changes requested):
All good, except some cache optimizations

brunobesson:
I do think, based on our needs, we could use alpine tag instead. When I first built the image, there weren't so many options.
Using the alpine tag would also simplify the update process.

florentcadot:
> I do think, based on our needs, we could use alpine tag instead. When I first built the image, there weren't so many options. Using the alpine tag would also simplify the update process.

I agree using the alpine tag would simplify the update process in the future.
I don't get why alpine would be better for the project need. Could you explain, please ?

brunobesson:
> I don't get why alpine would be better for the project need. Could you explain, please ?

I don't say it is better, it is however lighter, and the constraints that come with it do not apply to our usecase.

florentcadot:
> > I don't get why alpine would be better for the project need. Could you explain, please ?
> 
> I don't say it is better, it is however lighter, and the constraints that come with it do not apply to our usecase.

For the "lighter" part, not sure if you're talking about the openresty image, if yes:

openresty/openresty:alpine 42.08 MB
https://hub.docker.com/r/openresty/openresty/tags?page=1&name=alpine

openresty/openresty:bookworm   35.01 MB
https://hub.docker.com/r/openresty/openresty/tags?page=1&name=bookworm

So bookworm seems lighter, or maybe I missed something ? 
 
Anyway, thank you for taking time to reply and your review :smile: 

And let me know if you need a PR using alpine

brunobesson:
Alpine-apk is 9mb.
I haven't checked th base images for this. Anyway, bookworm is fine, but
we'll need to stay aware of debian updates.

Le jeu. 4 janv. 2024, 17:22, Florent Cadot ***@***.***> a
écrit :

brunobesson (approved):

Issue #3571, opened 2023-08-01
Bug: when searching on the map to recenter it automatically no action happens


anakreenskyrider:
**Describe the bug**
Bug: when searching on the map to recenter it automatically no action happens

**To Reproduce**

1. Go to map
2. search a place
3. validate or select it ith the mouse
4. it should recenter on the selected location
5. nothing happens :)

**Expected behavior**
4. it should recenter on the selected location

**Screenshots**
refer to : [https://forum.camptocamp.org/t/bug-centrage-de-la-carte/319573/4](https://forum.camptocamp.org/t/bug-centrage-de-la-carte/319573/4)

**Desktop (please complete the following information):**

- Device:PC
- OS: windows
- Browser chrome
- VersionVersion 115.0.5790.110 

**Additional context**
also reported for firefox

florentcadot:
Here is a PR trying to fix this issue: https://github.com/c2corg/c2c_ui/pull/3787

florentcadot:
Duplicated with https://github.com/c2corg/c2c_ui/issues/3568 I guess

PR #3787, opened 2024-01-06
fix: #3571 recenter on map search

florentcadot:
Here is a PR to fix https://github.com/c2corg/c2c_ui/issues/3571

Not sure why the condition `if(extent)` was necessary, the problem was (cf changes for reference):
 -  `let extent = feature.get('extent');` is truthy
 -  `extent = ol.proj.transformExtent(extent, 'EPSG:4326', 'EPSG:3857');` set `extent` to `[Infinity, Infinity, Infinity, Infinity]`
 -  So `this.view.fit(extent, { size: this.map.getSize(), maxZoom: 12 });` return an error
 
My guess is that `transformExtent` needs an extent parameter with the following type according to the declaration file:

````
/**
 * An array of numbers representing an extent: `[minx, miny, maxx, maxy]`.
 */
export type Extent = Array<number>;
````

...and I'm not sure the value passed to `transformExtent` has the right type (`[minx, maxx, miny, maxy]`) I guess.

### DIFF
--- a/src/components/map/OlMap.vue
+++ b/src/components/map/OlMap.vue
@@ -1170,18 +1170,10 @@ export default {
     // https://github.com/c2corg/v6_ui/blob/c9962a6c3bac0670eab732d563f9f480379f84d1/c2corg_ui/static/js/map/search.js#L194
     recenterOn(item) {
       const feature = geoJSONFormat.readFeature(item);
-      let extent = feature.get('extent');
       let coordinates = feature.getGeometry().flatCoordinates;
-
-      if (extent) {
-        extent = ol.proj.transformExtent(extent, 'EPSG:4326', 'EPSG:3857');
-        this.view.fit(extent, { size: this.map.getSize(), maxZoom: 12 });
-      } else {
-        coordinates = ol.proj.transform(coordinates, 'EPSG:4326', 'EPSG:3857');
-        this.view.setCenter(coordinates);
-        this.view.setZoom(16);
-      }
-
+      coordinates = ol.proj.transform(coordinates, 'EPSG:4326', 'EPSG:3857');
+      this.view.setCenter(coordinates);
+      this.view.setZoom(16);
       this.showRecenterOnPropositions = false;
     },
 

### CODE REVIEW

**Comment by eddy-geek** [comment]
thanks @florentcadot !

@c2corg/developers  can we merge this one? it's a very annoying bug.

I validated the change on my laptop (for the add route, add waypoint, and edit outing pages), if it helps.

**Comment by lbesson** [comment]
Hello,

It's been a long time I haven't touched ol, do you know if we are still able to retrieve the extent in another way?
Is the 16 zoom not also "breaking" cases where we have a geometry that is not  a point?

**Comment by eddy-geek** [comment]
yes "areas" is what I mean by "not a point" -- and again, "not  point" is arguably useless, and zoom 16 is much better than broken :wink: 

still, I had a look, but for now I can only confirm the weird behaviour of transformExtent explained by @florentcadot 

 

```js
//  add `debugger;` in the problematic vuejs code, then in the google chrome console type each line:

ol = _js_libs_ol__WEBPACK_IMPORTED_MODULE_29__.default;
// {Map: ƒ, View: ƒ, Feature: ƒ, Collection: ƒ, Overlay: ƒ, …}
coordinates
// (2) [12.674297, 42.6384261]
extent
// (4) [6.6272658, 47.0921462, 18.7844746, 35.2889616, __ob__: Observer]
 ol.proj.transform(coordinates, 'EPSG:4326', 'EPSG:3857');
// (2) [1410896.2882027149, 5257097.552514766]
ol.proj.transformExtent(extent, 'EPSG:4326', 'EPSG:3857');
// (4) [Infinity, Infinity, -Infinity, -Infinity]
```

eddy-geek:
Ok chatgpt helpfully pointed out

> It seems like there might be an issue with the extent array you're using as an argument for ol.proj.transformExtent. The array for extent should follow the order of [minx, miny, maxx, maxy] in EPSG:4326 coordinates.

and provided the correct code for my example (Italie)

```js
ol.proj.transformExtent([6.6272658, 35.2889616, 18.7844746, 47.0921462], 'EPSG:4326', 'EPSG:3857');
//(4) [737743.8542076768, 4203219.538521824, 2091078.1472911811, 5957127.668862533]
```

florentcadot:
I close this one as @eddy-geek has provided the correct solution here  : https://github.com/c2corg/c2c_ui/pull/3794

PR #3794, opened 2024-01-10
fix: #3571 recenter on map search. Supersedes #3787

eddy-geek:


lbesson:
I can't test sorry, but if it works, I think that is fine.

florentcadot:
I've just tested, works fine !

brunobesson (approved):

PR #3811, opened 2024-01-19
feat: allow visible password

xiaomutoun:


src/views/user/utils/FormField.vue
```diff
@@ -40,6 +40,14 @@
         class="input"
         :class="{ 'is-danger': hasError || errorMessage }"
       />
+      <span
+        v-if="type === 'password'"
+        class="icon is-small is-right"
+        @click="onMaskToggle"
+        style="pointer-events: initial"
+      >
```
brunobesson, line 48: Please use a class instead, e.g. 'is-clickable' helper from bulma css.

```suggestion
      <span
        v-if="type === 'password'"
        class="icon is-small is-right is-clickable"
        @click="onMaskToggle"
      >
```

src/views/user/utils/FormField.vue
```diff
@@ -40,6 +40,14 @@
...
+        <fa-icon :icon="unmasked ? 'eye-slash' : 'eye'" />
```
brunobesson, line 49: One could add a title to the input so that the action is explicit : `:title="$gettext(unmasked ? 'Hide password' : 'Show Password', 'Passord prompt')"`

brunobesson (changes requested):
Thanks for your work! And sorry for the delay before reviewing and answering.

This is a nice feature, I only have minor accessibility comments, easy to solve.

xiaomutoun:
The corrections are done.

brunobesson (approved):

Issue #3713, opened 2023-11-15
ToS acceptation


lbesson:
This is a follow up of the backend part: https://github.com/c2corg/v6_api/pull/1542

When the user logs in in the UI, and if the ToS are not accepted, a 403 will be returned with the body 'Terms of Service need to be accepted'

In that case, the UI should re-display the login form with a tickbox for accepting terms of conditions, and all 3 fields (login, email and tickbox) must be sent to the backend so that the ToS acceptation is stored in the db. Next logins will not get the 403.

Note: the backend part has not been merged yet. Please see with developers in order to enable this for demo if you start working on it, or run the branch of the backend locally.

florentcadot:
Here is one attempt to solve this issue: https://github.com/c2corg/c2c_ui/pull/3838

eddy-geek:
@c2corg/developers closing following the last load including #3838. Reopen if I'm wrong

PR #3838, opened 2024-02-11
feat: #3713 show ToS checkbox if ToS error on login

florentcadot:
One attempt to solve https://github.com/c2corg/c2c_ui/issues/3713

As written in , https://github.com/c2corg/v6_api/pull/1542 should be merged first.

image

## How to test:

- For the backend checkout to the branch used in this PR: https://github.com/c2corg/v6_api/pull/1542 and run the migrations
- Launch the web app and try to login with any existing users (username: user590, mdp: c2c for instance), it should de facto fail as no users has accepted the ToS.
- When it fail it should show the reason why it fail above the login form and if it's because the user didn't accept the ToS it should show a checkbox below the password field. If you check the box and press login it should successfully login

### DIFF
--- a/src/views/user/LoginView.vue
+++ b/src/views/user/LoginView.vue
@@ -23,6 +23,17 @@
         icon="key"
       />
 
+      <div class="field" v-if="showLoginTerms">
+        <div class="control">
+          <label class="checkbox">
+            <input type="checkbox" v-model="loginTermsAgreed" />
+            <span class="has-text-weight-bold" v-translate>
+              I have read and agree to the <router-link :to="termsRoute" v-translate>terms of service</router-link>.
+            </span>
+          </label>
+        </div>
+      </div>
+
       <div class="has-text-weight-bold is-centered has-text-centered mb-1" v-translate>
         Beware of case-sensitive login and password
       </div>
@@ -74,8 +85,9 @@
           <label class="checkbox">
             <input type="checkbox" v-model="termsAgreed" />
             &nbsp;
-            <span class="has-text-weight-bold" v-translate>I have read and agree to the terms of use</span>
-            (<router-link :to="{ name: 'article', params: { id: 106731 } }" v-translate>link</router-link>).
+            <span class="has-text-weight-bold" v-translate>
+              I have read and agree to the <router-link :to="termsRoute" v-translate>terms of service</router-link>.
+            </span>
           </label>
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


src/views/user/LoginView.vue
```diff
@@ -23,6 +23,16 @@
         icon="key"
       />
 
+      <div class="field">
+        <div class="control">
+          <label class="checkbox" v-if="showLoginTerms">
```
brunobesson, line 28: why is the `v-if` set there and not on line 26?

src/views/user/LoginView.vue
```diff
@@ -23,6 +23,16 @@
...
+            <input type="checkbox" v-model="loginTermsAgreed" />
+            <span class="has-text-weight-bold" v-translate>{{ termsMessage }}</span>
+            (<router-link :to="termsRoute" v-translate>link</router-link>).
```
brunobesson, line 31: link is not supposed to be transated?

src/views/user/LoginView.vue
```diff
@@ -225,10 +234,17 @@ export default {
       redirectionStillDone: false,
 
       promise: {},
+      loginTermsAgreed: undefined,
+      termsMessage: 'I have read and agree to the terms of use',
```
brunobesson, line 238: there is an inconsistency with the other denomination `terms of service`

src/views/user/LoginView.vue
```diff
@@ -313,8 +329,20 @@ export default {
...
+    },
+
     signin() {
-      this.promise = this.$user.signIn(this.username, this.password).then(this.onSuccessSigin);
+      this.promise = this.$user
+        .signIn(this.username, this.password, this.loginTermsAgreed)
+        .then(this.onSuccessSigin)
+        .catch((e) => {
+          this.isTermsError(e) ? (this.loginTermsAgreed = false) : (this.loginTermsAgreed = undefined);
+        });
```
brunobesson, line 345:
```suggestion
        .catch((e) => (this.loginTermsAgreed = !this.isTermsError(e)));
```

brunobesson, line 345:
```suggestion
        .catch((e) => (this.loginTermsAgreed = this.isTermsError(e) ? false : undefined));
```

brunobesson (approved):
Some minor comments, otherwise the PR looks really good. I haven't tested it yet however, for the same reasons explained in the PR for external resources in waypoints.

brunobesson (changes requested):
Some minor comments, otherwise the PR looks really good. I haven't tested it yet however, for the same reasons explained in the PR for external resources in waypoints.

src/views/user/LoginView.vue
florentcadot: What do you mean ? `link` will be translated here isn't it ?

brunobesson: sorry, I assume I misread. however, the name is not explicit enough for the translation system. Either rename it or add a context, please.

brunobesson (approved):

Issue #3697, opened 2023-11-03
Feature: add External Resources section on waypoints


BorutK:
Please add an **External resources** section on [waypoints](https://www.camptocamp.org/waypoints/add/fr)! 

In the case of routes (routes have such a section), this had permitted bots **not to correct that section** and to keep article and book titles in their original state. This is [particularly expected](https://www.camptocamp.org/articles/996571/fr/corrections-semi-automatiques-dans-le-topoguide#note-importante-a-lattention-des-dresseurs-de-bot) when it comes to the Camptocamp abbreviation of cardinal points, or titles of foreign guides (literature), links...

florentcadot:
Here are the front and back PR for this feature:
https://github.com/c2corg/c2c_ui/pull/3846
https://github.com/c2corg/v6_api/pull/1747

brunobesson:
API has been updated in production.

BorutK:
Bravo

________________________________

PR #3846, opened 2024-02-18
feat: #3697 add waypoints external resources section

florentcadot:
https://github.com/c2corg/c2c_ui/issues/3697

Associated backend PR:  https://github.com/c2corg/v6_api/pull/1747

# How to test:

Pull the associated backend branch `feature/add-waypoints-external-resources`
Run the added migration using docker-compose exec api .build/venv/bin/alembic upgrade head
 
Create a new waypoint and fill the `external_resources` field. Then update the external resources of this.
 
 You should see something like this:
 
Screenshot from 2024-02-18 11-03-40

image

### DIFF
--- a/src/js/constants/documentsProperties.json
+++ b/src/js/constants/documentsProperties.json
@@ -650,7 +650,8 @@
       { "id": "books", "properties": { "url": "b", "associationEditorOrder": 2 } },
       { "id": "images", "properties": { "associationEditorOrder": 4 } },
       { "id": "waypoints", "properties": { "url": "w", "associationEditorOrder": 1 } },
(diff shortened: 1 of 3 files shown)


brunobesson (approved):
It looks good to me thanks. I haven't tested it though because I couldn't manage to launch the API side on my laptop (work restrictions + I am not fluent with python).

IMHO, we can merge both parts (backend and frontend).

brunobesson:
API has been updated in production. We should test and merge if ok.

brunobesson:
I thus tested this PR, and it is working fine 👏

However, we should have a specific placeholder, different from that for routes.

image

PR #3868, opened 2024-03-06
feat: c2corg#846 colorize waypoint with transportation services

Nayor:


brunobesson:
Hello, thanks for you PR!

The problem is, the `public_transportation_rating` info is not retured by the API for lists yet. So I wonder how you managed to have green icons on the map?

Nayor:
Oh, true, I haven't noticed the waypoints schema was restricted when fetch in lists
I've sent a PR for the API too

eddy-geek:
linking the backend PR here for ref : ~https://github.com/c2corg/v6_api/pull/1759

Nayor:
> The problem is, the `public_transportation_rating` info is not retured by the API for lists yet. So I wonder how you managed to have green icons on the map?
@brunobesson Since the PR for the back has been merged, the information is now available in the API for lists 🙂
Is this PR ok then ?

brunobesson:
Yes it is, alas the backend is not deployed in prodution yet. We can merge however, it won't break if the backend is not up to date.

PR #3873, opened 2024-03-08
fix: c2corg#2313 api climbing error translation

Nayor:


brunobesson (changes requested):
Hi,

the i18n system we use is a bit more complex than this, sorry 🙁 

First, keys are automatically extracted from the code, as described here: https://github.com/c2corg/c2c_ui/wiki/i18n, and then hanled through the transifex online tool. What should be done here, because the key cannot be inferred from the code, is to add an hint to the tool (simiar to what has been done in `src\views\wiki\edition\utils\document-edition-view-mixin.js`, but e.g. in the association tool). The translations will then be updated on transifex (you could request access if you wish) and PR generated automatically through a github action.

brunobesson:
> Are you thinking about displaying this error when an climbing indoor waypoint is selected by the association tool ?

Indeed.

> Wouldn't be the best to not display at all this kind of waypoint in the input suggestions ?

This should be the case already, but you never know :)

PR #4565, opened 2026-05-12
feat: add geo: URI link on document map box

Sixtedsm:
## What

Adds a button next to the GPX/KML download buttons on the document map box. On mobile, tapping it opens the document's location in the user's preferred map app (Google Maps, OsmAnd, Maps.me, Organic Maps, Plans, etc.) via a `geo:` URI ([RFC 5870](https://datatracker.ietf.org/doc/html/rfc5870)).

## Why

Requested on the forum by community members: https://forum.camptocamp.org/t/application-camptocamp-cahier-des-charges-a-elaborer-avec-vous/308792

Currently, to navigate to a waypoint or trailhead from camptocamp.org on a phone, users have to copy coordinates manually or download the GPX and import it elsewhere. This adds a one-tap path that integrates with whatever map app the user already uses.

## Scope

- One file modified: `src/views/document/utils/boxes/MapBox.vue`
- ~25 lines added, no dependencies, no API changes
- Button only renders on mobile (`$screen.isMobile`) and only when the document has a Point geometry
- Falls back gracefully (returns `null`, button is hidden) if geometry is missing or malformed
- Uses `ol.proj.toLonLat` to convert from the internal EPSG:3857 projection to WGS84 expected by the `geo:` URI scheme

## How to test

1. `npm run serve`
2. Open DevTools, toggle device toolbar (`Ctrl+Shift+M`), pick a mobile device profile
3. Navigate to any waypoint, e.g. http://localhost:8080/waypoints
4. The new pin icon appears next to the GPX / KML buttons under the map
5. Inspect the `<a>` element — the `href` is `geo:LAT,LON?q=LAT,LON` with 6-decimal precision (≈10 cm)
6. On a real mobile device, tapping the button opens the OS's "Open with..." dialog with installed map apps

## Notes

The `geo:` URI scheme has no handler on desktop browsers — the button is intentionally hidden there (`v-if="... && $screen.isMobile"`) to keep the UI clean.

## Summary by CodeRabbit

* **New Features**
  * Mobile users can now open location data directly in their device's default map application via a new "Open this location in your map app" button, available when location information is present.

[Review Change Stack](https://app.coderabbit.ai/change-stack/c2corg/c2c_ui/pull/4565)

### DIFF
--- a/src/views/document/utils/boxes/MapBox.vue
+++ b/src/views/document/utils/boxes/MapBox.vue
@@ -39,6 +39,14 @@
     </div>
 
     <div class="buttons is-centered">
+      <a
+        v-if="geoUri && $screen.isMobile"
+        :href="geoUri"
+        class="button is-primary is-small"
+        :title="$gettext('Open this location in your map app')"
+      >
+        <fa-icon icon="map-marker-alt" />
+      </a>
       <button v-if="showDownloadTraceButtons" class="button is-primary is-small" @click="downloadGpx">GPX</button>
       <button v-if="showDownloadTraceButtons" class="button is-primary is-small" @click="downloadKml">KML</button>
       <button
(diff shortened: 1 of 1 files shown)


eddy-geek:
thanks for the PR :)
the rabbit comment on malformed URI seems not applicable as I guess we validate `geom` on save.
please add a screenshot here for ref of the expected behaviour

PR #4572, opened 2026-09-12
fix(security): cors, GitHub Actions, PR Preview, HTML sanitization

Nayor:
## Summary by CodeRabbit

- **Security**
  - Added stronger browser security protections, including content restrictions, clickjacking protection, and safer referrer handling.
  - Sanitized forum, wiki, helper, and map-provider content before display to reduce cross-site scripting risks.
  - Restricted embedded video capabilities for safer playback.

- **New Features**
  - Added protected pull-request preview deployments for eligible contributions.

- **Documentation**
  - Documented the pull-request preview process and required repository configuration.

- **Maintenance**
  - Updated vulnerability patch policy and strengthened automated workflow permissions.


flob38:
Ca parle chinois pour moi.
Je ne vais pas être d'une grande utilité pour commenter les modifs.

Juste 1 question : est-ce que le sanitize conserve les images et liens ?

Nayor:
> Ca parle chinois pour moi. Je ne vais pas être d'une grande utilité pour commenter les modifs.

Merci pour ton retour ! Principalement des modifs de sécu
Seul changement un peu majeur ; il faudra une validation côté GitHub pour déployer la preview d'une PR front

src/components/generics/Markdown.vue
```diff
@@ -77,7 +78,9 @@ export default {
         }
       };
 
-      container.innerHTML = this.content;
+      // this.content comes from wiki articles converted from user-authored markdown:
+      // sanitize it before injecting it in the DOM to prevent stored XSS.
+      container.innerHTML = sanitizeRichHtml(this.content);
```
eddy-geek, line 83: Will this break automatically turning link text into `<a href>` ?

eddy-geek:
Can we double check that all the sanitization doesn't result in loss of usability? Beginners don't know about markdown urls syntax, and I'm not sure how turning http into links is less secure.

src/components/generics/Markdown.vue
Nayor: Verified with a real DOMPurify run: <a href="..."> passes through both sanitizeHtml and sanitizeRichHtml completely untouched
<a>/href are in DOMPurify's default allow-list and are never stripped. Sanitization only runs after markdown→HTML conversion (done server-side), so it can't affect how links get generated in the first place.

Nayor:
> Can we double check that all the sanitization doesn't result in loss of usability? Beginners don't know about markdown urls syntax, and I'm not sure how turning http into links is less secure.

Checked the backend markdown parser (c2corg_api/markdown/__init__.py) ; it has no autolink extension enabled today, with or without this change. So pasting a bare URL without [text](url) syntax doesn't become clickable currently, and this PR doesn't make that worse or better, it's an orthogonal, pre-existing parser behavior, not something the sanitizer restricts. If we want that for beginners, it's a backend change (enable an autolink extension), and it wouldn't be "less secure": DOMPurify already allows <a href="http://..."> unconditionally, so an autolinked plain URL would sanitize exactly the same as a manually-linked one.

Separately worth noting: the wiki content this sanitizer processes already goes through server-side bleach cleaning before reaching the frontend, so this DOMPurify pass is mostly defense-in-depth there. It's not redundant everywhere though — forum posts (Discourse) never pass through that backend cleaner, so for that content path this is the only sanitization c2c_ui controls.

